PR #224, opened 2016-12-06
API refactoring to allow management of non-Smalltalk code

demarey:


demarey:
This PR is pending since about 2 months. What do you want to do with it? There are new commits and the merge become harder day after day

estebanlm:
meh... I will need to check at it :)

estebanlm:
@npasserini I think I'm ready to integrate this... what do you say?

PR #229, opened 2016-12-12
add "gitlocal" capability

estebanlm:
with this PR we can execute scrips of this type: 

```Smalltalk
Metacello new
	repository: 'gitlocal://Users/esteban/Dev/Repository/voyage/mc';
	baseline: 'Voyage';
	load.
```


astares:
Is "gitlocal" an official protocol? Why not use file:// insead?

I tried file:// and it worked for me.

PR #262, opened 2017-01-25
Multi remotes

estebanlm:
IMO, this is ready to integrate


estebanlm:
this tests should pass when latest vm is promoted to stable. 
also, this merge cannot wait more... is too big.

PR #268, opened 2017-02-02
first shot to plugins (and GitHub PR as test case)

estebanlm:
and I do this PR using my thing :)


BaselineOfIceberg.package/.filetree
```diff
@@ -1,5 +1,4 @@
 {
-	"separateMethodMetaAndSource" : false,
 	"noMethodMetaData" : true,
-	"useCypressPropertiesFile" : true
-}
\ No newline at end of file
+	"separateMethodMetaAndSource" : false,
+	"useCypressPropertiesFile" : true }
```
jecisc: It could be cool if:
- The exporter could not rewrite metadata without changes
- The exporter could not change the order of metadata between two export.

I am the only one to be bothered by this?

BaselineOfIceberg.package/BaselineOfIceberg.class/properties.json
```diff
@@ -1,11 +1,14 @@
...
+	"instvars" : [
+		 ],
 	"name" : "BaselineOfIceberg",
-	"type" : "normal"
-}
\ No newline at end of file
+	"pools" : [
+		 ],
+	"super" : "BaselineOf",
+	"type" : "normal" }
```
jecisc: Same here, the exporter produce a lot of noise for nothing :(

jecisc: The files are exported from FileTree right? So If I open an issue it is on FileTree?

estebanlm: yep

jecisc: Ty, I opened an issue (https://github.com/dalehenrich/filetree/issues/214)

Iceberg-Plugin.package/IceGitHubPullRequestModel.class/instance/obtainAvailableBranchNames.st
```diff
@@ -0,0 +1,8 @@
+private
+obtainAvailableBranchNames
+	^ ((self repository allBranches
```
npasserini: I think this is weird, because the same list of branch names is used for both base and head repositories, and there is no warranties that branches might exists in both sides. Isn't it possible to change the set of branches accordingly to the selected repository?

estebanlm: not really at that stage (also is not that what github does). What github does is to allow you to put a branch name if you are not convinced with the ones that he offers you. 
We can add that later, too

PR #271, opened 2017-02-03
a fix for how to show in/out commits on multi-remotes

### LINKED ISSUE #270: using multi-remotes does not show correctly in/out elements
so, we introduced the possibility of have multiple remotes. 
we also introduced the capability of define "default pulling/pushing remote" for the tool, different to the one that is defined as default (who is origin/branch usually)... the problem is that the browser does not show correctly this definitions hence introducing a lot of confusion.

estebanlm:
fixes #270


estebanlm:
@npasserini do not like this solution, so I will close it :)

PR #272, opened 2017-02-03
show compact changes on trees

### LINKED ISSUE #269: showing compacted version of changes?
So... when managing large projects, I find annoying the diff panel tree that shows all packages, even those that do not have changes... in that view in fact I want to focus on the modified packages and for example on pharo-core I have to scroll down always to see anything (I have like 300 packages, and sometimes a one-line change). 

What do you think? It has sense?

estebanlm:
fixes #269


estebanlm:
the PR depends on fix for #270 who is being reworked, I will submit a new version that does not depends on it.

Issue #81, opened 2016-07-04
Review cloneRepository


npasserini:
Currently, if you try to clone an scp url, cloneRepository will change the url and convert it into an https url. This is done because of a bug between libgit2/libssh2 which keeps ssh authentication from working and
cloning with https doesn't require credentials. 

Yet, current tests do not use this behavior so it is not clear that it works. Moreover, I tested that ssh does not work with clone, but this implementation suggests that it might work for other commands. "

estebanlm:
fix is integrated

PR #276, opened 2017-02-09
enable use of ssh

### LINKED ISSUE #275: enable use of ssh-agent
the use of SSH it is disabled now (only usable ay now is HTTPS), because libgit2 was not working some time ago, but now we need to implement it.

estebanlm:
As it says, this PR enabled the use of ssh and ssh-agent on iceberg. This should fix some concerns around (also, is a lot easier to use than provide credentials by hand).

this fix closes #275


estebanlm:
this PR closes #81 too

estebanlm:
so... any reason why I shouldn't merge this?

PR #280, opened 2017-02-14
in diff trees show just packages with changes

### LINKED ISSUE #269: showing compacted version of changes?

estebanlm:
closes #269


Iceberg-UI.package/IceDiffChangeTreeBuilder.class/instance/elements.st
```diff
@@ -1,3 +1,3 @@
 accessing
 elements 
-	^ self diff elements
\ No newline at end of file
+	^ self diff elements select: #hasChanges
```
estebanlm: this is the real change

estebanlm:
this is annoying when it decides the property files needs to be updated too :S

estebanlm:
so... any complains ?

PR #285, opened 2017-02-21
Multi remotes

npasserini:


Iceberg.package/IceBranch.class/class/named.local.inRepository..st
```diff
@@ -1,4 +1,5 @@
 instance creation
 named: branchName local: isLocal inRepository: repo
+	self haltIf: [ branchName includes: Character space ].
```
estebanlm: this needs to be here?

estebanlm:
I do not see any relevant change here :P

estebanlm:
I think this became old

PR #303, opened 2017-03-21
create #compatibleUsername who will answer FirstnameLastname instead …

### LINKED ISSUE #302: when loading a package, author name of method timestamps is wrong
when installing a method, it needs to be

FirstnameLastname mm/dd/yyyy hh:mm

what actually happens is, since in git most usernames are Firstname Lastname, we have a broken timestamp.

estebanlm:
…'Firstname Lastname'

fixes #302


estebanlm:
I didn't just pushed because I want a second opinion :)

estebanlm:
ok... no comments means "go ahead" :)

Issue #292, opened 2017-03-07
error when switching branch


demarey:
When you try to switch to another branch from the Branches tab (main iceberg panel), you get a red morph because iceberg thows an error saying that `Reference 'HEAD' is not a local branch`.
Closing and reopening Iceberg UI will throw again this error.

demarey:
to reproduce:
- load iceberg:dev-04,
- open iceberg and clone a repository (give iceberg url)
- try to switch to dev-0.4 branch

demarey:
IceLibgitLocalRepository>>#upstream tries to send the message `#upstream` to `repo head` that is  a reference to 'HEAD' and not a reference to a local branch as expected by [git_branch_upstream](https://libgit2.github.com/libgit2/#HEAD/group/branch/git_branch_upstream)

demarey:
maybe we should use `git_branch_lookup`?

demarey:
duplicate of #250

demarey:
In fact, this problem appears because one try to witch to a remote branch, without having a local branch for it in the local repository.
The solution would be to first create the local branch before doing the checkout when one want to switch to a remote branch

estebanlm:
fixed

PR #311, opened 2017-03-24
fix Issue #292

demarey:
avoid errors when switching to a remote branch by creating a local branch if needed.
Detached head mode is not yet supported.


estebanlm:
again, I will accept it, but I do not like this properties explosion... we need to check what's going on :)

jecisc:
@estebanlm I opened an issue some time ago on FileTree about that issue: https://github.com/dalehenrich/filetree/issues/214

Issue #282, opened 2017-02-16
Cannot forget a repository if the local git directory does not exist anymore


demarey:
scenario:
- clone a new repository in Iceberg
- save and quit
- remove the pharo-local repository
- open the image
- try to forget the repository in Iceberg UI => LGit_GIT_ENOTFOUND: Failed to resolve path ...

Uko:
➕1️⃣

PR #315, opened 2017-03-28
fix issue #282

demarey:


Iceberg.package/IceRepository.class/instance/origin.st
```diff
@@ -1,3 +1,9 @@
 accessing remotes
 origin
-	^ self backend origin
\ No newline at end of file
+	^ [ self backend origin ]
+		on: LGit_GIT_ENOTFOUND 
+		do: [ :error | "if local repository is deleted, LGit_GIT_ENOTFOUND will be thrown. We already know the origin"
+				{ pullRemote . pushRemote }
```
estebanlm: this needs to be `self pushRemote. self pullRemote`, not variables directly, because they can be nil (not initialised) and tnen the rest of the verification will have a DNU :P

Iceberg.package/IceRepository.class/instance/origin.st
```diff
@@ -1,3 +1,9 @@
...
+					detect: [ :remote | remote remoteName = #origin ]
```
estebanlm: `remote isOrigin`, not hardcoded constants, please :)

demarey:
it is a fix for issue #282 
I made a mistake on previous commit log. The rationale is explained here: https://github.com/npasserini/iceberg/issues/282. In short, if the repository is missing, you cannot ask the repository its origin ...
I saw #isOrigin but commited the wrong version because I had to rebuild an image (still this cache problem in iceberg ui)

estebanlm:
ahh, now I understand :)

PR #324, opened 2017-04-07
Swapped colors, so additions in Iceberg are green and removals are red

Uko:
Diff in Iceberg is super annoying because the colors are messed up


Iceberg-UI.package/SwappedColorDiffMorph.class/instance/additionColor.st
```diff
@@ -0,0 +1,3 @@
+accessing colors
+additionColor
+	^ super removalColor
```
estebanlm: ok, this scares me :)

PR #325, opened 2017-04-10
lazySnapshots

pavel-krivanek:
create snapshots of only dirty MC pacakges for faster commits

### DIFF
--- a/Iceberg-UI.package/IceRepositoriesBrowser.class/instance/addPackageToRepository..st
+++ b/Iceberg-UI.package/IceRepositoriesBrowser.class/instance/addPackageToRepository..st
@@ -9,4 +9,6 @@ addPackageToRepository: repo
 	options := (MCWorkingCopy allManagers difference: repo workingCopies).
 
 	(self class chooseFrom: options displaying: descriptionBlock) 
-		ifNotNil: [ :workingCopy | repo addPackageNamed: workingCopy packageName ]
+		ifNotNil: [ :workingCopy | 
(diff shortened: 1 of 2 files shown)


estebanlm:
this was not enough to fix the "added package" problem, isn't?

Iceberg-UI.package/IceRepositoriesBrowser.class/instance/addPackageToRepository..st
```diff
@@ -9,4 +9,6 @@ addPackageToRepository: repo
 	options := (MCWorkingCopy allManagers difference: repo workingCopies).
 
 	(self class chooseFrom: options displaying: descriptionBlock) 
-		ifNotNil: [ :workingCopy | repo addPackageNamed: workingCopy packageName ]
\ No newline at end of file
+		ifNotNil: [ :workingCopy | 
+			workingCopy modified: true.
```
estebanlm: I would prefer this logic extracted to a method other than start putting this inside the block.

pavel-krivanek:
no, but because the commit issue of a new package  is common with the mainstream version, it should be fixed separately.

estebanlm:
anyway, according to Nico this was not enough either because this is used to also compare any snapshot with current sources... @npasserini ?

pavel-krivanek:
what about to provide a new UI command, something like Synchronize dirty packages

estebanlm:
this needs to be reimplemented :P

PR #327, opened 2017-04-10
display a progress bar while computing versions

pavel-krivanek:


estebanlm:
sorry, this implementation does not works, at all: 

- it mixes UI with backend logic.
- it will not "inform" user: it will show a dialog that freezes everything until done. Yes is better than nothing, but this can be done better, through notifications/callbacks/etc.

PR #336, opened 2017-04-18
Guess code subdirectory when importing repo

peteruhnak:
* When importing a local repo, guess the code subdirectory if there is .filetree file present.
  * This is a convenience fix to save some typing
* Fixed button label for import local repo

p.s. I am not sure to which branch I should direct the PR... dev-0.4 is apparently the working branch... so here?

### DIFF
--- a/Iceberg-Libgit.package/IceLibgitLocalRepository.class/instance/cloneRepositoryFrom.branch..st
+++ b/Iceberg-Libgit.package/IceLibgitLocalRepository.class/instance/cloneRepositoryFrom.branch..st
@@ -6,7 +6,7 @@ cloneRepositoryFrom: aRemote branch: aBranchName
 	LGitGlobal runSequence: [ 
 		| repo cloneOptions |
 		repo := LGitRepository on: self location.
-		cloneOptions := LGitCloneOptions withCredentialsProvider: (IceCredentialsProvider default).
+		cloneOptions := LGitCloneOptions withCredentialsProvider: IceCredentialsProvider default.
 		
 		repo clone: url options: cloneOptions.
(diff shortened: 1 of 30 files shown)


Iceberg-UI.package/IceCreateRepositoryModel.class/instance/initializePresenter.st
```diff
@@ -1,3 +1,7 @@
 initialization
 initializePresenter
-	createButton action: [ self createRepository ]
\ No newline at end of file
+	createButton action: [ self createRepository ].
+	localDirectoryLocation locationInput
+		whenTextChanged: [ :newText | 
+			self location isNotNil & subdirectory text isEmpty
+				ifTrue: [ subdirectory text: (self guessSubdirectoryIn: self location) ] ]
```
estebanlm: some remarks: 
- I do not like logic so complex in an event block, I'd prefer to extract a method there.
- this: `self location isNotNil & subdirectory text isEmpty`
is in general incorrect (unnecessary). It should be better like thos: 
`self location notNil and: [ subdirectory text isEmpty ]`

peteruhnak:
Extracted (also I split the decision into two to avoid parenthesis and extra ifs)

Btw. I wanted to write test for this, but there are no tests in Iceberg, or am I missing something?

estebanlm:
yes there are tests. 
If you install `#development` it will install everything to test...
I think some of them are broken right now... there is a problem with VM version on smalltalk-ci and our capability of test things in linux... but we need to work on restore it (just... I didn't have the time).

peteruhnak:
I was fixing my branches, so this was automatically closed... I'll reopen it later with tests...

PR #337, opened 2017-04-24
Updated travis config

peteruhnak:
* Changed Pharo-6 to Pharo-alpha, because stable VM doesn't support iceberg's git (note that we moved SmalltalkCI's alpha from stable vm to latest vm just now)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,19 +6,18 @@ os:
 - osx
 
 smalltalk:
-- Pharo-6.0
-- Pharo-5.0
+- Pharo-alpha
 
 before_install:
 # Installing the dependency until it is included in the VM
 - if [ $TRAVIS_OS_NAME == "osx" ] ; then brew install libgcrypt --universal; fi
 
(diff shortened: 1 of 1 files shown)


estebanlm:
I do not think it is the good solution for testing with vm latest.
it is possible that you want to test an alpha image with a stable vm. 
it is possible that you want to test both.
I would have add vm: `#stable`, vm: `#latest` as options. And also an option for threaded vm.
I guess this is an opinion to send to smalltalk-ci, not here :)

peteruhnak:
I've changed it to vm-latest, because Iceberg/libgit doesn't support the current stable VM; it fails to load libgit and libssh, because the file structure of the vm is different

estebanlm:
by the way... you made the PR to master, not dev-0.5

estebanlm:
and yes, using latest it was ok... I was just arguing that use Pharo-alpha to provide a latest vm was not the best choice, IMO.

PR #338, opened 2017-04-24

peteruhnak:
* Changed Pharo-6 to Pharo-alpha, because stable VM doesn't support
iceberg's git

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,19 +6,18 @@ os:
 - osx
 
 smalltalk:
-- Pharo-6.0
-- Pharo-5.0
+- Pharo-alpha
 
 before_install:
 # Installing the dependency until it is included in the VM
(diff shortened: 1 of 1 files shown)


peteruhnak:
> and yes, using latest it was ok... I was just arguing that use Pharo-alpha to provide a latest vm was not the best choice, IMO.

Fabio has in his plans to split image and vm configuration, but that requires a bit more changes.

peteruhnak:
Now what I don't understand is why the load fails here, and not in the source repo:

source repo: loading ok, 18 tests ok, 55 tests error ( https://travis-ci.org/peteruhnak/iceberg/jobs/225239554 )
here repo: loading fail `#withSqueakLineEndings` ( https://travis-ci.org/npasserini/iceberg/jobs/225240382 )

estebanlm:
wow... no clue :P

PR #339, opened 2017-04-24

peteruhnak:
* Changed Pharo-6 to Pharo-alpha, because stable VM doesn't support
iceberg's git

second attempt, based from dev-0.5 branch (although it shouldn't make a difference)


peteruhnak:
Ok, load somehow magically fixed itself. Now only the tests are failing.

PR #342, opened 2017-05-08
Use pharo-vcs instead of npasserini

jecisc:


estebanlm:
oops, I missed that one :)

Issue #347, opened 2017-05-11
IceCommitModel redefine the methods of the Trait it use


jecisc:
IceCommitModel contains two methods of the Trait TIceTool but they are recompiled into the class. Maybe this was cause by a bug with trait?

jecisc:
Done in df86f59ebf097bad5751de4b856fdfa4e2ce2de6

Issue #258, opened 2017-01-17
Iceberg should remember last setting of "Upload changes on commit"


astares:
Let's assume that you have a github project locally cloned and from time to time you commit your work.
After doing some coding you usually synchronize to commit your changes with Iceberg.

Let's also assume that you work offline - not connected to the internet (for instance in a train)
and later push your branch to the origin.

In the sync view there is a checkbox "Upload changes on commit?" - so if you work OFFLINE you can unselect this box.  If you commit often you have to unselect it very often. Doing this again and again 
before each commit is annoying. Iceberg should remember the previous setting of this checkbox.

astares:
Eclipse solved this by having two buttons "Commit" and "Commit and Push".

See http://wiki.eclipse.org/images/f/ff/Egit-3.1-StagingView.png

jecisc:
I also think that two buttons is better than the checkbox.

jecisc:
@estebanlm Can be closed

PR #348, opened 2017-05-11
Replace "Push" checkbox by "Commit and Push" button

jecisc:
- Remove the setting to remember if the commits should be pushed
- Remove the checkbox to know if a commit should be pushed
- Add a new button "Commit and Push"
- Remove methods duplicated from a Trait


jecisc:
I am not sure this PR will be accepted but personally I find it easier to use to get two buttons instead of a button and a checkbox. One click less when we want to commit without pushing and one less setting to manage. 
I wait your though about this PR :)

jecisc:
Related: https://github.com/pharo-vcs/iceberg/issues/258 and https://github.com/pharo-vcs/iceberg/issues/347

PR #354, opened 2017-05-21
url parsing fix

theseion:
minor fix to http URL parsing that allows '.' to be part of the last path segment


estebanlm:
hi @theseion 
can I ask you to make the PR against dev-0.4?
while dev-0.5 is what is coming, I'm doing some huge changes there and because of that it will take some more time to hit master :S

theseion:
Replaced by PR #363 on dev-0.4

PR #355, opened 2017-05-21
Update IceNetworkRemote>>userName

Hockenberry:
The method 

IceNetworkRemote>>userName
	^user ifNil: [ String empty ] ifNotNil: [ user ].

 should be 

IceNetworkRemote>>userName
	^user ifNil: [ String empty ] ifNotNil: ''.

### DIFF
--- a/Iceberg.package/IceNetworkRemote.class/instance/userName.st
+++ b/Iceberg.package/IceNetworkRemote.class/instance/userName.st
@@ -1,3 +1,3 @@
 accessing
 userName
-	^user ifNil: [ String empty ] ifNotNil: ''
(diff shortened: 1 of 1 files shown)


estebanlm:
this is incorrect: 
- `String empty` is the same as `''`
- in this case userName is *always* an empty string, which is bad.

Hockenberry:
Sorry for my bad english and rushing.

It must read:

The method
IceNetworkRemote>>userName
^user ifNil: [ String empty ] ifNotNil: ''.

should be

IceNetworkRemote>>userName
^user ifNil: [ String empty ] ifNotNil: [ user ].

Hockenberry:
Is this incorrect too? See patch ^^

PR #363, opened 2017-05-25
URL parsing fix

theseion:
minor fix to http URL parsing that allows '.' to be part of the last path segment


estebanlm:
why this is again here?

theseion:
> On 12 Jun 2017, at 13:10, Esteban Lorenzano <notifications@github.com> wrote:
> 
> why this is again here?
> 

Um... what is your question exactly? I made two separate fixes to the regex, if that is what you're asking.

estebanlm:
ahhh :)

Issue #349, opened 2017-05-11
Sort the branches by remote


jecisc:
Later it would be cool to sort the branches by remote. 

Here is a screen of another tool doing it and it's much simpler to understand: [https://blog.axosoft.com/wp-content/uploads/2015/10/6-add-remote-full.png](https://blog.axosoft.com/wp-content/uploads/2015/10/6-add-remote-full.png)

jecisc:
Done in: a77d0d332df29fc948fbfe194d9310ac54f943b8

PR #391, opened 2017-07-09
cf_groupBranchesByRemote

jecisc:
This group the branches by remotes in the branches view.

Issue related: https://github.com/pharo-vcs/iceberg/issues/349

Known bug: The elements of the tree should be openned but for now the FastTreeTable of Glamour cannot do it. This is a Glamour bug and not an Icebeng one. See: https://pharo.fogbugz.com/f/cases/20229/GLMFastTreeTablePresentation-does-not-expand-the-elements-if-we-ask-it


jecisc:
Hum… This has too many changes not related to my work. I'll try to do a PR without that.

PR #392, opened 2017-07-09
cf_groupBranchesByRemote

jecisc:
This group the branches by remotes in the branches view.

Issue related: https://github.com/pharo-vcs/iceberg/issues/349


jecisc:
This is better :)

Issue #393, opened 2017-07-12
Add filter field for package tab of repository


dionisiydk:

jecisc:
How do you want to filter? If the name begins with the pattern? If the name includes the pattern?

dionisiydk:
I want same filter like in MonticelloBrowser. It uses "includes the pattern" logic

jecisc:
PR#394 waiting

jecisc:
@estebanlm Done.

PR #394, opened 2017-07-12
cf_addFilterToPackageView

jecisc:
Add a filter to the package view. Probably usefull for projects like Pharo. 

Issue related: https://github.com/pharo-vcs/iceberg/issues/393

### DIFF
--- a/Iceberg-UI.package/IceRepositoriesBrowser.class/instance/composePackagesListIn..st
+++ b/Iceberg-UI.package/IceRepositoriesBrowser.class/instance/composePackagesListIn..st
@@ -14,6 +14,7 @@ composePackagesListIn: composite
 		addSelectionAction: self unloadAndRemovePackageAction;
 		addSelectionAction: self removePackageAction;
 		addSelectionAction: self removePackageFromDiskAction;
+		enableFilter: [ :packageModel :pattern | packageModel package name includesSubstring: pattern caseSensitive: false ];
 		"Plugin actions"
 		dynamicActionsOnSelection: [ :presentation | 
 			presentation entity pluginPackageActionsFor: presentation selection entity ];


Iceberg-UI.package/IceRepositoriesBrowser.class/instance/composePackagesListIn..st
```diff
@@ -14,6 +14,7 @@ composePackagesListIn: composite
 		addSelectionAction: self unloadAndRemovePackageAction;
 		addSelectionAction: self removePackageAction;
 		addSelectionAction: self removePackageFromDiskAction;
+		enableFilter: [ :packageModel :pattern | packageModel package name includesSubstring: pattern ];
```
dionisiydk: Could your use instead:
 includesSubstring: pattern caseSensitive: false

jecisc: Done

Iceberg-UI.package/IceRepositoriesBrowser.class/instance/composePackagesListIn..st
```diff
@@ -14,6 +14,7 @@ composePackagesListIn: composite
 		addSelectionAction: self unloadAndRemovePackageAction;
 		addSelectionAction: self removePackageAction;
 		addSelectionAction: self removePackageFromDiskAction;
+		enableFilter: [ :packageModel :pattern | packageModel package name includesSubstring: pattern caseSensitive: false ];
```
estebanlm: using `model package name` doesn't feels right. Can you use `model packageName` ?
(I know, it does not exists... you need to add it)

Issue #390, opened 2017-07-08
Add shortcut to open Iceberg


jecisc:
Most of Pharo's tools can be open by a shortcut as "Meta + o, Meta + aLetter". 

I think "Meta + o, Meta + i" is not used. Maybe Iceberg could use it.

jecisc:
Corrected

PR #395, opened 2017-07-12
cf_registerIcebergInShortcuts

jecisc:
Add shortcut to open Iceberg. (meta o + meta i).

Issue related: https://github.com/pharo-vcs/iceberg/issues/390


jecisc:
Hum… There is unwanted files here… Weird.

PR #413, opened 2017-07-18
dev-0.5

noha:
Fixed multi-level submodule directories


noha:
There are too many modifications. Need to do again

Issue #421, opened 2017-07-23
Guess subdirectory during local repository creation


jecisc:
This is just an idea but maybe we could guess the code subdirectory of a project when we add a local subdirectory. 

When we set the folder we can scan the directories to find the one containing .filetree file.

estebanlm:
merged

PR #422, opened 2017-07-23
Guess the subdirectory while importing a local repository.

jecisc:
See: https://github.com/pharo-vcs/iceberg/issues/421


estebanlm:
I don't like the "chooseHook" name, is not coherent with our method names. 
instead, I would use "onChoose" (and chooseBlock as instance variable).

jecisc:
@estebanlm Done.

estebanlm:
ok, just another remark: the common convention name for block closure parameters is "aBlock", not "aBlockClosure" :) 
I will accept it with it since is not so bad, I'm telling you for the future :)

Issue #431, opened 2017-08-13
UX improvements: default pull/push remote menu


jecisc:
Currently, if I understand correctly, if a remote is a default pull/push remote, the option in the contextual menu is disabled. I think it should not be disabled and if the remote is already the default one we should show an inform to the user with "X is already the default Y remote". When a menu item is disabled we don't know the reason and I was confused at first to see that.

jecisc:
@estebanlm I don't know how to show titles depending on the state of the model and the selected object in GLMGenericAction. Do you know how we can do this?

estebanlm:
meh, maybe it is not possible "as is" (I think we need to change glamour to support it). 
I will check when I'm back from holidays.

guillep:
Not valid anymore. We came back to a system of upstream (as in git) and explicit selection of pull and push branches/remotes.

PR #433, opened 2017-08-13
Improve ux of default pull push remote menu

jecisc:
Now "Make default pull/push remote" is not disabled in the menus but inform the user of why it does nothing.

Close issue: https://github.com/pharo-vcs/iceberg/issues/431


estebanlm:
I do not like this change, I still prefer to have them disabled (maybe changing the text of the disabled menu will fit better)

jecisc:
Ok, I originaly did this because this is a rule that I saw in all UX recommendations: never disable a menu entry.

(For example: https://www.joelonsoftware.com/2008/07/01/dont-hide-or-disable-menu-items/)

estebanlm:
Nevertheless most other guidelines do it (like osx, windows, etc.)

PR #451, opened 2017-08-26
450 cherry picking from multiple packages mixes up methods

### LINKED ISSUE #450: Cherry picking from multiple packages mixes up methods

theseion:
use package specific collection to create patch for cherry pick to prevent method mixups.
Fixes #450


theseion:
@estebanlm This fix is critical for the integration process

Iceberg.package/IceDiff.class/instance/repository..st
```diff
@@ -2,4 +2,4 @@ accessing
 repository: anObject
 	repository := anObject.
 	repository announcer weak
-		when: IceCommited send: #refresh to: self
\ No newline at end of file
+		subscribe: IceCommited send: #refresh to: self
```
jecisc: I think this change is not wanted.

theseion: sh***! thanks!

PR #452, opened 2017-08-26
450 cherry picking from multiple packages mixes up methods

### LINKED ISSUE #450: Cherry picking from multiple packages mixes up methods

theseion:
use package specific collection to create patch for cherry pick to prevent method mixups.
Fixes #450


estebanlm:
oops. This looks like the right thing :)

Issue #383, opened 2017-07-03
Unpushed commit color is hard to read with the dark theme


jecisc:
When I have unpushed commit one Iceberg, the repo is Blue. But it is really hard to read.

jecisc:
Culprit is: IceAbstractModel that does not rely on the Theme.

jecisc:
The problem is that I see nothing usable in the Theme :(

jecisc:
https://pharo.fogbugz.com/f/cases/20249/Add-some-colors-to-the-themes

jecisc:
I proposed to add some colors to Pharo 7 themes. When this will be done we will be able to use them. We can add a Iceberg-Pharo6 package with these methods to not break compatibility.

estebanlm:
done

PR #456, opened 2017-08-30
fix(ui): do not use hardcoded colors for Iceberg

jecisc:
Since the color used where added in Pharo 7, add a package to load them in Pharo 6 too. 

Close https://github.com/pharo-vcs/iceberg/issues/383


jecisc:
Wait this before integrating: https://github.com/pharo-project/pharo/pull/231

jecisc:
Mergeable :)

PR #462, opened 2017-09-13
typo

StephanEggermont:


estebanlm:
we use dev-0.5 as dev branch. 
master is stable (I know, you have opinions with that... but for now is like that :) )

PR #469, opened 2017-09-25
Added a noop flush for get ( as per MCRepository>>#flushForScriptGet )

peteruhnak:
Added noop flush (same as MCRepository) to allow `get`ting baselines.

An argument could be made that flushing should actually pull the latest version (just like it does in github+filetree), however considering we are typically using the local repo it would be a volatile change. So flushing (pulling) should be (imho) done manually via Iceberg.


estebanlm:
heh... same fix is already in dev-0.6 :)

peteruhnak:
Ah, excellent. Closing this one then. :)

PR #472, opened 2017-09-29
[ENH]: Fix #471 - Setting for Default Code Subdirectory

### LINKED ISSUE #471: Setting for Default Code Subfolder
Fix incoming…

seandenigris:
- Example: `IceRepository defaultCodeSubdirectory: aString "e.g. 'src'"`
- Default to 'source' per http://forum.world.st/Beacon-moved-to-github-tp4953881p4954017.html
- Also available via Settings Browers


jecisc:
I am not really in favor of `source` but since it is a personnal taste I can accept it, but anyway it should be "sources" instead of "source" IMO.

estebanlm:
I do not see "source" but "repository" ?

estebanlm:
I'm not really fan of "source" either... because I'm old, I prefer src... but community seems to prefer source so I will accept PR now (just I will merge it to dev-0.6 because I will release this version soon)

seandenigris:
@estebanlm I was on the fence about source vs src as well. I didn't have a strong feeling either way, so I went with the the way the ML thread was leaning. Yet, to be honest, it wasn't exactly a broad consensus, just the last two or three comments there. I figured most of us would set it to whatever we prefer, so it would really only matter for encouraging newbies to use the convention - if there even is one!

estebanlm:
truth is, most people will use what *we* provide (and they will declare it "standard", so later will not be easy to change ;) but well... we cannot always reach a consensus. 
in other hand, I moved my own projects and pharo projects to src, so who knows... in any case, we can change it later, after some iterations.

PR #473, opened 2017-10-04
Url handling

zecke:


estebanlm:
Sorry to arrive late to this. 
What's this PR about? (I'm out of context)

zecke:
Well. It is partially about using RegExps to parse URLs and fall-outs from it. I started with importing a project that has ssh://git@host:port/path as origin and then found a related ticket.

* git:// URLs can not be parsed (add it)
* file:// URLs actually can be absolute paths too. E.g. git clone somedir otherdir (detect more as file urls)
* Use ZnUrl to parse file and http but not git.. as git:git@foo:path is not a valid URL.. (or URI)

estebanlm:
ok, let's see :)

estebanlm:
agggg. It had to be on dev-0.6 :(

PR #474, opened 2017-10-04
$/ in a branch name is common. Pass indication of isRemote down

zecke:
I use some semantics in branch names. E.g. fixes/proj-number or 
feature/feature-name.  Or shared projects use user/their-work so make
it work. Pass name instead of baseName to get the full branch (the
refs/heads/ will be removed by other code) and pass the indication of
local/remote down to libgit.


zecke:
This requires https://github.com/pharo-vcs/libgit2-pharo-bindings/pull/15 of libgit2-pharo-bindings...

PR #476, opened 2017-10-06
default code directory is empty for the moment

estebanlm:


estebanlm:
yeah, I do not like source. So for the moment I let it empty

Issue #481, opened 2017-10-10
iceberg fails to load private bitbucket repositories


estebanlm:
an url style: estebanlm@bitbucket.org:estebanlm/private-project.git

should be possible.

PR #485, opened 2017-10-11
Correct usage of bitbucket typed repository

noha:
Iceberg repository should return MCBitbucketRepository instead of MCGithubRepository


Iceberg.package/MCBitbucketRepository.extension/instance/getOrCreateIcebergRepository.st
```diff
@@ -0,0 +1,15 @@
+*Iceberg-Adapters
+getOrCreateIcebergRepository
```
estebanlm: shouldn't this be moved to its parent?

noha: I was asking that myself but wasn't sure

estebanlm: so?

noha: I can move it up

estebanlm:
btw, this will close #481 ?

Iceberg.package/MCBitbucketRepository.extension/instance/getOrCreateIcebergRepository.st
noha: done

noha:
I'm not sure what #481 is about exactly but it sounds that it is the same case. It is a two-fold fix. The first is in libgit accepting paths with leading slashes and thus enable multi-level dirs. The second fix is that a bitbucket instance is used instead of github. This way you can use bitbucket://user/project in metacello

estebanlm:
yes it is

PR #532, opened 2017-12-08
20831 Bad gitlocal: reference causes infinite loop and out of memory

akgrant43:
Attempting to load a project with metacello and a gitlocal: reference that isn't actually a git repository (not bare repository) causes an infinite loop and ultimately out of memory error.

The error is in IceGitLocalRepositoryType>>splitRootAndSubdirectoryFromLocation which tests for root isNil, instead of root isRoot (the parent of the root directory is the root directory, not nil).


estebanlm:
fixes #534

akgrant43:
Hi Esteban,

> just, contributions go in the current development branch (dev-0.6). I changed the base

Sorry about that.  I was looking for a 'development' branch and my eyes glossed over the dev-xxx branches.

Thanks,
Alistair

PR #536, opened 2017-12-12
Update README.md

apocalypsemystic:
- Added clarification on instructions I made mistakes on when attempting to follow along.

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,13 +104,13 @@ Metacello new
 
   - Local directory: You can just leave the default location for a start.
 
-  - Code subdirectory: You should provide the name of a subdirectory inside of your repository inside which your code lies. Usual options are `src`, `mc` or `repository`. Leaving it empty means that the code is in the root of your repository, this is a valid choice too.
+  - Code subdirectory: You should provide the name of a subdirectory inside of your repository inside which your code lies. Usual options are `src`, `mc` or `repository`. Leaving it empty means that the code is in the root of your repository, this is a valid choice too. Iceberg will *not* create the code direcotry for you if it does not exist, so if you are cloning a brand new repository with no pre-made code directories in it, leave this blank.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -104,13 +104,13 @@ Metacello new
 
   - Local directory: You can just leave the default location for a start.
 
-  - Code subdirectory: You should provide the name of a subdirectory inside of your repository inside which your code lies. Usual options are `src`, `mc` or `repository`. Leaving it empty means that the code is in the root of your repository, this is a valid choice too.
+  - Code subdirectory: You should provide the name of a subdirectory inside of your repository inside which your code lies. Usual options are `src`, `mc` or `repository`. Leaving it empty means that the code is in the root of your repository, this is a valid choice too. Iceberg will *not* create the code direcotry for you if it does not exist, so if you are cloning a brand new repository with no pre-made code directories in it, leave this blank.
```
jecisc: direcotry  should be directory

guillep:
Hi @apocalypsemystic, could you check if those two "details" are still problems? Otherwise, I'd prefer to fix them instead of put some workaround in the doc.

apocalypsemystic:
Sure. Which version of Pharo/Iceberg should I test with?

guillep:
The one in latest Pharo 7 should be good enough. However, we are probably releasing version (0.7.3) today, so if you do it in a couple of hours/this evening, you will be using latest release.

apocalypsemystic:
I'm not really sure they're "problems" per se, it just depends on what you're trying to do.

Best I can tell, if you enter a "code/" source path while cloning a new repo from Github, and that path does not exist, it throws the error DirectoryDoesNotExist. If you're cloning a well-established project, that might be what you want, but if you're cloning a brand new repo, like I was when I wrote the comment originally, it might be confusing, since I was expecting it to "set up" my empty repo by creating the directory. Maybe a "create if does not exist" checkbox would make sense.

Issue #574, opened 2018-02-09
Failure to parse Gitolite-style owner-less URLs


martinmcclure:
See also Pharo case 20895.
https://pharo.fogbugz.com/f/cases/20895/IceScpRemote-parseUrl-fails-to-parse-some-valid-urls

IceScpRemote>>parseUrl does not work for owner-less URLs of the type used by, for instance, Gitolite. These URLs typically look like either
gitolite3@server.bogus.edu:reponame
or
ssh://gitolite3@server.bogus.edu/reponame

I am preparing a proposed fix, will issue a pull request when ready.

martinmcclure:
Pull request made.

PR #575, opened 2018-02-09
Issue574

martinmcclure:
Tests and proposed fix for issue 574, adding support for owner-less URLs of the type used by Gitolite.


estebanlm:
fix #574

PR #587, opened 2018-03-31
Revert changes only if revert operation is available.

### LINKED ISSUE #288: MessageNotUnderstood: IceClassChangeSet>>operation While reverting changes
Today I try to revert some changes from the synch view but I got: MessageNotUnderstood: IceClassChangeSet>>operation

It happen when you try to revert the deletion of a class at least.

peteruhnak:
Revert changes only if revert operation is available.
Add option to recursively revert changeset.

fixes #288

### DIFF
--- a/Iceberg-UI.package/IceDiffChangeTreeBuilder.class/instance/buildOn..st
+++ b/Iceberg-UI.package/IceDiffChangeTreeBuilder.class/instance/buildOn..st
@@ -11,6 +11,7 @@ buildOn: aPresenter
 		addSelectionAction: self browseAction;
 		addSelectionAction: self browseReferencesAction;
 		addSelectionAction: self revertChangeAction;
+		addSelectionAction: self revertChangesRecursivelyAction;
 		onChangeOfPort: #entity act: [ :presentation | 
 			self diff ifNotNil: [  
 				self diff announcer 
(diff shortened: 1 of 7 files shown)


estebanlm:
I could merge this, but I'm not sure it has sense, since we are about to release 0.7 which will change all this radically (yes, you are in the list of "required testers" for that release, just is still not ready for testing :P).

estebanlm:
(btw, the new version does reverse of method, class and package correctly)

peteruhnak:
Does your approach also support revert vs revert recursively? Regardless, if you already resolved this in 0.7 then we can close this PR

guillep:
Hi @peteruhnak, Can you check if what is there in 0.7 is ok for you? I don't think the PR is valid anymore but the issue is.

Actually in version 0.7 reverts work recursively to all childs. The only case where a "revert only this one but not its childs" are class changes that modify instance variables or change the class tag. However, if you revert an instance variable you may break the class' methods :). So my question is: is it worth it?

estebanlm:
hi @peteruhnak 
this PR does not has sense anymore. 
if what is provided now is not enough, we need to work out a different solution.

peteruhnak:
> However, if you revert an instance variable you may break the class' methods :)
@guillep you can make the same arguments about methods... if you revert a method you may break all the callers.

Furthermore unlike removing methods, accidentally removing instance variables will not actually break anything (unless you already have existing instances), as Pharo will happily continue to store value to removed variables.

But this is just a minor issue, the main problem was error on reverting, and not being able to revert parent with all its subelements, both of which are now working. 👍

Issue #634, opened 2018-04-13
Travis is not building Pull Requests


guillep:
Currently the build requires an ssh key to clone test repositories.
The repository contains an encrypted key and they variables for its decryption are safe encrypted variables in travis, which are ignored for security reasons in the build.

However, in general a key should not be needed to clone...

guillep:
cloning from github requires to have a github account if using ssh. I think for now not using ssh in tests is not a bad solution...

guillep:
So I've preferred to:
  - use the already existing parameterizable tests to run metacello integration tests using ssh and https tests locally
  - then in CI ssh tests are skipped for now

This allows us at least to test metacello integration. Testing ssh connection is another story...

guillep:
Build is ok https://travis-ci.org/pharo-vcs/iceberg/builds/366043523

PR #635, opened 2018-04-13
Make PR builds run

guillep:
Try to simplify the ssh usage until tests for PRs are green.

Otherwise we will have to slip tests between "those that can run in PRs and those that cannot", but that's not a satisfactory solution.


guillep:
Fixes issue #634

Issue #665, opened 2018-04-20
Sensibility of package filter seems odd


jecisc:
**OS**

Windows 10

**Pharo**

60540

**Iceberg**

0.7.2

**Problem**

I tried to add three packages to my repository:
- ChartJs-Core
- ChartJs-Demo
- ChartJs-Library

So I tried to filter with "Chart" but I got 0 result.

Then I tried with "chart" and I got results. That's odd.

![2018-04-21_00-53-44](https://user-images.githubusercontent.com/9519971/39076907-8b972a30-44fe-11e8-9c65-251ef823385b.gif)

guillep:
We've seen this during the sprint, yes

PR #700, opened 2018-04-24
cf_correctAddPackagesFilter

jecisc:
When adding a package, if we filter the list, use #asLowercase on both package name and pattern from the user.

Fixes https://github.com/pharo-vcs/iceberg/issues/665


guillep:
Thanks for the tests!

Fixes #665

Issue #619, opened 2018-04-11
IcePharoPlugin crash in Pharo 6.1


jecisc:
**OS:** Windows 10

**Pharo:** #60540

**Iceberg version:** v0.7.1

**What I did:**

My settings add by default Pharo to my image via this script:

```Smalltalk
"Add Pharo by default"
	(IceRepositoryCreator new
		location: StartupPreferencesLoader preferencesGeneralFolder / 'GitRepositories' / 'pharo';
		subdirectory: 'src';
		createRepository) register.
```

**Problem:**

The problem is that the IcePharoPlugin needs the SystemVersion>>#commitHash of the Pharo version to work and this information is only in Pharo 7.

I know that it does not make much sense to use the IcePharoPlugin in Pharo 6.1 but some people can have that on their settings. So maybe the IcePharoPlugin could check the Pharo version and work only in the current image is > Pharo 6.

estebanlm:
but why do you add pharo-repository/pharo in pharo 6.1? (it should work in P7, but not in P6.1)

jecisc:
Because I just added the Iceberg stuff in my scripts without thinking about that :)

I'll change my scripts but maybe there are other persons with this problem and disabling the IcePharoPlugin in Pharo 6 seems not hard to do so I proposed. If this issue is just closed for "you should not load pharo repository in Pharo 6.1" it's fine with me too :)

PR #713, opened 2018-04-25
IcePharoPlugin should not care of Pharo 6 since it'll just bug.

jecisc:
Fixes https://github.com/pharo-vcs/iceberg/issues/619


guillep:
The failing tests are not related.

Thanks!

Issue #697, opened 2018-04-24
Recompile after update


jecisc:
I tried to install dev-0.7 to the latest Pharo 7 and I got problems because of Obsolete classes in Iceberg. 

Maybe the script to update Iceberg should end with `OpalCompiler recompileAll`?

What do you think?

guillep:
That's not an iceberg problem. That's a problem from Calypso's plugin...
And that's a Pharo problem in general also:

https://pharo.fogbugz.com/f/cases/21519/RemoveFromSystem-does-not-work-well-with-Undeclareds

PR #715, opened 2018-04-26
Add a comment in the README about obsolete classes

jecisc:
Fixes https://github.com/pharo-vcs/iceberg/issues/697


jecisc:
@tesonep I think it was better to merge it in master since this is the page the users see when coming on the repository and it's a problem with the current stable version that is corrected via documentation.

guillep:
It was actually me that merged. I impersonated @tesonep :).

We should not merge to master directly. This will be introduced in next's week release, so it will not stay there for ever ;)

Issue #669, opened 2018-04-23
Cannot load project in Pharo 6.1 with Iceberg dev-0.7


jecisc:
**OS**

Windows 10

**Pharo**
60540

**Iceberg** 

dev-0.7 (2376c1b544a0da23538293913f52fea177c9879c)

**What I did**

- Download a new Pharo 6.1 image
- Open it without launching the settings
- Install Iceberg dev-0.7
- Launch settings (in my case they set the default folder, my ssh key and register a project)
- Try to install prismJs

```Smalltalk
Metacello new
	githubUser: 'DuneSt' project: 'PrismCodeDisplayer' commitish: 'development' path: 'src';
	baseline: 'PrismCodeDisplayer';
	onUpgrade: [ :e | e useIncoming ];
	onWarningLog;
	load
```

Current PrismJS SHA on the development branch: 56c14637c84a08301046ea2dca144f05792ec144

**Result**

I get an error:

```
SmallInteger(Object)>>altDNU:
SmallInteger(Object)>>doesNotUnderstand: #isSeparator
STONReader>>consumeWhitespace
STONReader>>next
STON class>>fromStream:
[ :stream | STON fromStream: stream ] in IceRepositoryProperties class>>fromCommitish: in Block: [ :stream | STON fromStream: stream ]
[ aBlock value: stream ] in FileReference(AbstractFileReference)>>readStreamDo: in Block: [ aBlock value: stream ]
BlockClosure>>ensure:
FileReference(AbstractFileReference)>>readStreamDo:
IceRepositoryProperties class>>fromCommitish:
IceGitCommit>>properties
IceGitCommit>>writerClass
IceGitCommit>>fetchPackageNames
IceGitCommit>>packageNames
IceGitCommit(IceCommit)>>packages
IceWorkingCopy>>initializePackagesFromRepository
IceRepositoryCreator>>addLocalRepository
IceRepositoryCreator>>addLocalRepositoryAndCheckout
IceMetacelloDuplicatedNotification>>useIncoming
[ :e | e useIncoming ] in UndefinedObject>>DoIt in Block: [ :e | e useIncoming ]
BlockClosure>>cull:
BlockClosure>>cull:cull:
BlockClosure>>cull:cull:cull:
MetacelloScriptEngine>>handleUpgrade:
```

*Note:* In the method `fromCommitish:`, if I try to inspect `ref` I get either an error about ZnBufferStream not indexable or a VM crash

jecisc:
[...]

PR #727, opened 2018-04-28
Adapt properties reading of Tonel projects for Pharo 6.1.

jecisc:
Fixes https://github.com/pharo-vcs/iceberg/issues/669

PLEASE REVIEW! Since I'm not that familiar with the streams maybe there is a better solution.

Also, It would be cool if it could make it for release v0.7.4 because I'm stuck with Iceberg v0.7.2 to work on my projects.


guillep:
This should not break anything in Pharo7, so it's ok.

jecisc:
Cool :) I'm eager to try v0.7.4.

Thanks for the fast integration!

Issue #725, opened 2018-04-28
Remove and clean old UI


guillep:
Issue coming from: #549.

The old UI is still there. We need to clean it up carefully to not break anything.

jecisc:
Some notes:
- `SwappedColorDiffMorph` is used by the new UI and should be moved to Iceberg-TipUI
- `IceGitHubPullRequestBrowser` inherits from the old UI
- When the old UI will be removed you can remove `ManifestIceberg >> ruleRBReturnsBooleanAndOtherRuleV1FalsePositive`

jecisc:
If I have some more notes I'll just update the previous comment to not flood the subject.

PR #734, opened 2018-04-28
Begin to remove old UI.

jecisc:
Part of https://github.com/pharo-vcs/iceberg/issues/725

**This is far from being done** but it removes a lot of unused classes and moves some stuff used by the new UI to its package.


guillep:
I had to revert it because it was pointing to master...

jecisc:
Oh :( Sorry about that.

Issue #735, opened 2018-04-28
Issue while registering a new project on Pharo 6.1


jecisc:
**OS**
Windows 10

**Pharo**

Latest update: #60540

**Iceberg**

v0.7.4

**Report**

While executing my settings containing:

```
(IceRepositoryCreator new
		location: StartupPreferencesLoader preferencesGeneralFolder;
		subdirectory: 'settings';
		createRepository) register
```

(which contains the clone of this repo: https://github.com/jecisc/pharo-scripts)

Cause this error:

```
UndefinedObject(Object)>>doesNotUnderstand: #nextChildNodeStartingFrom:suchThat:ifNone:
IceNode>>firstNodeSuchThat:ifNone:
IceNode>>firstNodeSuchThat:
[ tree
	firstNodeSuchThat: [ :definition | self isCodeSubdirectory: definition ] ] in IceDiff>>codeSubdirectoryNode in Block: [ tree...
BlockClosure>>cull:
ByteString(Collection)>>ifNotEmpty:ifEmpty:
IceDiff>>codeSubdirectoryNode
IceDiff>>includesPackageNamed:
[ :each | each beDirty: (diff includesPackageNamed: each name) ] in IceWorkingCopy>>adoptCommit: in Block: [ :each | each beDirty: (diff includesPackageNamed...etc...
Array(SequenceableCollection)>>do:
IceWorkingCopy>>adoptCommit:
IceWorkingCopy>>initializePackagesFromRepository
IceRepositoryCreator>>addLocalRepository
[ self validate.
self isCloning
	ifTrue: [ self cloneRepository ]
	ifFalse: [ self addLocalRepository ] ] in IceRepositoryCreator>>createRepository in Block: [ self validate....
BlockClosure>>on:do:
IceRepositoryCreator>>createRepository

```

jecisc:
Not reproducible on P7.

jecisc:
I located the problem.

In:
```
isCodeSubdirectory: aNode
	
	^ aNode path pathString = self repository subdirectory
```

In P6.1  `aNode path pathString` is "/settings" and `self repository subdirectory` is "settings".

In P7 they're both "settings"

jecisc:
Problem is that Path class>>#withAll: changed between P6.1 and P7

PR #737, opened 2018-04-28
Correct detection of subdirectiory in Pharo 6.1

jecisc:
The representation of Paths changed between Pharo 6.1 and Pharo 7 and the detection of subdirectories is broken in Pharo 6.1 because of that.

Fixes: https://github.com/pharo-vcs/iceberg/issues/735

PLEASE READ THE CHANGES BEFORE INTEGRATING IT!!

I'm not familiar with Paths and maybe there is a better correction to do.

Also maybe it would be cool to test this but I'm not sure how to set up a repository and I don't have the time to do more for now :(

### DIFF
--- a/Iceberg.package/IceDiff.class/instance/isCodeSubdirectory..st
+++ b/Iceberg.package/IceDiff.class/instance/isCodeSubdirectory..st
@@ -1,4 +1,15 @@
 testing
 isCodeSubdirectory: aNode
+	| nodePathString |
+	self flag: #clean. "
+	This is a hack for Pharo 6.1 because representation of paths changed between Pharo 6.1 and Pharo 7.
+	When Pharo 6.1 will not be maintaned anymose this method can be changed for: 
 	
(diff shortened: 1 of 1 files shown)


guillep:
Yeh, I'll check it carefully. I'd like to see if the extra slash comes from an extra empty child at the beginning of the path, because of the usage of an absolute path instead of a relative one, or some other issue.

Ideally, I'd like to avoid patching code in iceberg. We should use compatible APIs if possible, otherwise a forward-compatibility package.

jecisc:
I agree. This is the reason I asked a careful review. This fix works but I'm sure there is better to do and I did not had the time to check how paths works. But at least I can use this patch in the meanwhile to use Iceberg with my projects.

guillep:
I've pushed a fix in 05d1589. It's a fix in the same spirit but I think it's a bit cleaner ^^

jecisc:
@guillep I agree, I prefer your solution.

Issue #739, opened 2018-04-29
Initial focus should be in "FogBugz issue" field, also rename the label


astares:
- open Iceberg
- select Pharo -> Create new branch for issue
- in the dialog the cursor is in the second edit field (Title) but should be in the first one (FogBugz issue) so one can enter an issue number and the title is grabbed from the issue tracking tool

Ideally we would also rename the label into just "Pharo Issue Number" because
FogBugz is now known as Manuscript (fogbugz.com redirects to manuscript.com)

jecisc:
I don't understand why the issue number field is not selected first...

In the code there is:
```
self focusOrder
		add: issueNumberText;
		add: issueText
```

So I assume it should be selected first. (But I don't know spec that much...)

jecisc:
Ok, the problem comes from the fact that this Presenters inherits from another one defining a focus order but all the content of the superclass does not seem to be used.

As the `branchInputText`

guillep:
Should be fixed with #740 for next release

PR #740, opened 2018-04-29
Fogbugz panel: Improve label, focus order and layout

jecisc:
- Rename label from FogBugz issue to Pharo issue Number
- The focus order now works right
- Redefine the extent of the window

Fixes https://github.com/pharo-vcs/iceberg/issues/739


jecisc:
I extracted the duplicated parts

guillep:
Tx @jecisc :)

astares:
Thanks!

guillep:
In the latest version the focus is not working... At least I cannot see it because Tab does not change focus at all when the focus is on the list on the left.

PR #766, opened 2018-05-04
Add copy SHA command to history window.

jecisc:
Fixes https://github.com/pharo-vcs/iceberg/issues/765


jecisc:
@guillep Updated.

PR #779, opened 2018-05-11
Make tests work on 64 bits

guillep:
Try latest commit of ossubprocess. WIP for #761


guillep:
Tests on 64 bits seem to work with that version

PR #792, opened 2018-05-16
Clean and move IceAskForPlaintextCredentialsModel

jecisc:
- Remove two unused variables
- Move the class to Iceberg-Plugin-Github since it's the only user
- Rename the class to add "GitHub" in the name

Maybe later this could be use for other plugins (Bitbucket/Gitlab) and we will need a "Iceberg-Plugin-AbstractHosting" package or something like that? But for now it's the only user.


guillep:
Part of #725

Issue #827, opened 2018-05-30
Add package dialog has some glitches


guillep:
- it shows all packages in the system, even those that are already tracked by the repository
- the OK button is enabled even if nothing was selected

PR #845, opened 2018-06-11
Improve add packages dialog

jecisc:
- Exclude packages already in the repository
- Disable accept button when no item is selected
- Rename "ok" button to "Add"

Fixes https://github.com/pharo-vcs/iceberg/issues/827


jecisc:
Error seems to be a random failure of OSSubprocess

guillep:
Looks like. I'm rerunning the two failing jobs and I'll integrate it later. Thanks!

PR #846, opened 2018-06-11
useCommander063

dionisiydk:
new Commander with deprecated renamed classes

### DIFF
--- a/BaselineOfIceberg.package/BaselineOfIceberg.class/instance/commander..st
+++ b/BaselineOfIceberg.package/BaselineOfIceberg.class/instance/commander..st
@@ -4,5 +4,5 @@ commander: spec
 		baseline: 'Commander' 
 		with: [ 
 			spec 
-				repository: 'github://pharo-ide/Commander:v0.6.2';
+				repository: 'github://pharo-ide/Commander:v0.6.3';
 				loads: #('Core' 'AllActivators' 'Commander-SpecSupport') ]


dionisiydk:
It should work fine with current Calypso version. 
But small testing is needed to approve that shortcuts and menu continue work in general (not every item)

guillep:
I don't understand.

- Is this PR ready for integration?
- who should do a "Small test"? Didn't you do it when you applied the change?
- More general question: How can we properly test Commander commands?

dionisiydk:
Hi @guillep. 
Possible issues are related to deprecated classes, how I managed them.
I tested the scenario when I just load new Commander into fresh image.
But I can't check how it will work when Iceberg is loaded from scratch. This is what I call small test.

dionisiydk:
About testing. There are two parts:
1) Main command logic is implemented in #execute method. So in test we can simply create command instance, execute it and check the result. It does not depends on any context where command can be used.
2) Testing of command initialization in particular context instance. 
Here we need to prepare context instance and call implemented methods from the tests.  It can be not easy to do because in preparation methods we use UI to request additional information. Also context can require many dependant objects to simulate real scenario.
So this part is an issue for testing and we need find a way to improve it.

guillep:
But then I still do not understand if this is safe to merge...

dionisiydk:
Yes, it is safe

пн, 18 июня 2018 г., 12:30 Guille Polito <notifications@github.com>:

Issue #854, opened 2018-06-13
Error in History window


jecisc:
In the latest Pharo 7 I tried to open the history window of RBParseTreeRewriter>>visitCascadeNode: and I got this error: `Instance of IceLogVersion did not understand #tagNames`

Stack:

```
IceLogVersion(Object)>>doesNotUnderstand: #tagNames
IceTipVersionModel(IceTipCommitModel)>>info
IceTipCachedModel>>forwardMessage:
IceTipCachedModel>>doesNotUnderstand: #info
IceTipExtendedVersionBrowser(IceTipVersionHistoryBrowser)>>refreshCommitInfo
IceTipExtendedVersionBrowser(IceTipVersionHistoryBrowser)>>showInfoTabs
IceTipExtendedVersionBrowser(IceTipVersionHistoryBrowser)>>refreshCommitTabPanel
IceTipExtendedVersionBrowser(IceTipVersionHistoryBrowser)>>refresh
IceTipExtendedVersionBrowser(IceTipVersionHistoryBrowser)>>initializeWidgets
IceTipExtendedVersionBrowser(ComposablePresenter)>>initializePrivateHooks
IceTipExtendedVersionBrowser(ComposablePresenter)>>initialize
IceTipExtendedVersionBrowser(IceTipBrowser)>>initialize
IceTipExtendedVersionBrowser class(ComposablePresenter class)>>on:
IceTipExtendedVersionBrowser class(IceTipVersionHistoryBrowser class)>>onRepository:method:
IceTipExtendedVersionBrowser class>>repository:method:
ClyIcebergShowMethodVersionCommand>>execute
ClyMethodContextOfFullBrowser(ClySystemBrowserContext)>>executeCommand:by:
[ self prepareCommandForExecution.
context executeCommand: command by: self.
self applyCommandResult ] in CmdCommandActivator>>executeCommand in Block: [ self prepareCommandForExecution....
BlockClosure>>on:do:
CmdCommandActivator>>executeCommand
[ aCommandActivator executeCommand ] in ClyIcebergShowMethodVersionCommand(CmdCommand)>>buildBrowserTabActionsUsing: in Block: [ aCommandActivator executeCommand ]
[ anAction action value ] in TabActionButtonMorph>>forAction: in Block: [ anAction action value ]
TabActionButtonMorph(MultistateButtonMorph)>>mouseUp:
```

PR #855, opened 2018-06-13
addMissingMethods

jecisc:
Add missing methods. Fixes https://github.com/pharo-vcs/iceberg/issues/854


jecisc:
Since there is no dev-1.1 branch I do it on dev-1.0. 

I can change the destination later if needed.

PR #867, opened 2018-06-19
Introduce first version of Projects

guillep:
- Add project file support
- Detecting project file
- Add a repair action to solve it


scripts/preLoading.st
```diff
@@ -18,7 +18,8 @@ MetacelloPharoPlatform select.
     'Iceberg-TipUI'
     'Iceberg-Plugin-Pharo' 
     'Iceberg-Plugin-Metacello' 
-    'Iceberg-Plugin-GitHub' 
+    'Iceberg-Plugin-GitHub'
+    'Iceberg-Plugin-Migration'
```
jecisc: This should also be in the update script of the README IMO. It will be needed after the integration of this code in Pharo.

guillep: Yup. Not for Pharo6, but yes for Pharo7.

PR #895, opened 2018-06-26
Fixes case when DELETE is returning a 204 No Content, for example usi…

### LINKED ISSUE #894: GitHubAPI fails when the API responds with a 204 No Content
I'm using the API to remove labels in my projects, and it fails because `delete:` expects a JSON content as response, and it's not always the case. For labels, for example, the API will return a 204 No Content when it's successful.

gcotelli:
…ng the API to delete labels in a project.

Fixes #894 .

I have reviewed the senders of `delete:` in the GitHubAPI and no one uses the result.

### DIFF
--- a/Iceberg-Plugin-GitHub.package/IceGitHubAPI.class/instance/delete..st
+++ b/Iceberg-Plugin-GitHub.package/IceGitHubAPI.class/instance/delete..st
@@ -1,6 +1,6 @@
 private requesting
 delete: aString 
-	^ self jsonContentsWithValidationDo: [
+	^ self contentsWithValidationDo: [
 		(self newRequestTo: aString)
 			delete;
 			response ]


gcotelli:
You can try it in a new testing repo using:
```smalltalk
| repo owner api labels |
repo := 'REPO'.
owner := 'OWNER'.
api := IceGitHubAPI new 
	credentials: (IcePlaintextCredentials new username: 'USER';password: 'PASSWORD';yourself).
labels := api get: ('repos/<1s>/<2s>/labels' expandMacrosWith: owner with: repo).
labels collect: [ :label | api delete: ('/repos/<1s>/<2s>/labels/<3s>' expandMacrosWith: owner with: repo with: (label at: #name) urlEncoded)  ].
```

guillep:
Thanks!

PR #909, opened 2018-06-29
Move wiki files to a docs directory (so we can accept pull requests) …

macta:
…along with some minor grammatical changes and a new entry for recovering code.

As suggested by Esteban to move away from the wiki.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,8 +3,7 @@ Iceberg is a set of tools that allows one to handle git repositories directly fr
 
 Please be aware that this is still an experimental tool, so you have to be careful. But we already have a few early adopters, so you are invited to try it and provide feedback.
 
-To better understand Iceberg (or even this documentation), I recommend to read the wiki pages which explain (parts of) the [Iceberg](../../wiki/Iceberg-glossary) and [Git](../../wiki/Some-keys-to-understand-Git-nomenclature) terminology.
-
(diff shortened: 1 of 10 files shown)


cdlm:
Still, the wiki is nice to browse on github. I think github wikis have their own git repo, but I should check if that's still the case. If it is, then the wiki content could be managed as a branch of the main repo; we'd have to synchronize it ourselves I guess, but then people could browse it online…

macta:
This is the annoying thing - it’s a second class repo so you can’t fork and PR - wasted hours on it already and found lots of similar complaints - it’s totally stupid.

cdlm:
What I mean is, you could have a branch in the 1st class repo that contains the same thing as the 2nd class one. I believe that would allow PRs, and synchronizing with the wiki repo would be normal push/pull, we just have to remember to do it.

Alternate solution: use `git subrepo` to synchronize the `docs/` directory with the wiki repo.

macta:
Damien - that looks interesting, I’ll give it a shot. Still don’t understand why github made this all so complicated in the first place (why did it even need to be a separate repo for a wiki) - what a disaster.

guillep:
The automatic upgrade could be done on the CI.

Let's integrate this, and work on the other solution afterwards.

PR #922, opened 2018-07-15
Cookbook: Add "Changing Last Commit Author"

seandenigris:


guillep:
I'm not completely sure that Iceberg should have a plain git documentation/cookbook... But it's ok for now.

PR #942, opened 2018-08-02
Fix #940

### LINKED ISSUE #940: Installation in new Pharo should also bootstrap pharo repository
It should call 

BaselineOfIDE new addPharoProjectToIceberg 

Otherwise we wouldn't know if it is failing or not.

tesonep:
Fix #940


tesonep:
Soy un bobo, it should not point to master!

estebanlm:
bobo x 2, you can change the target of the PR pressing edit :)

PR #971, opened 2018-08-10
issue/967

### LINKED ISSUE #967: Merging a Commit that does not have Smalltalk Code
When there is need of creating a merge commit and the incoming commit does not include changes in Smalltalk code the commit is not created, a "Nothing to commit" message is shown and the working copy stays in "In Merge" state (having two reference commits)

tesonep:
When producing a merge, sometimes Github produces a merge commit that is not necessary. 
This merge commit includes two commits that are also included in the chain of ancestors of the head.

The merge should always create the commit with the two parents.
Fix #967

### DIFF
--- a/Iceberg.package/IceMerge.class/instance/updateHead.st
+++ b/Iceberg.package/IceMerge.class/instance/updateHead.st
@@ -17,5 +17,5 @@ updateHead
 				setMergeStateBetweenCommits:
 					{self leftCommit.
 					mergeCommit}.
-			repository workingCopy commitChanges: self withMessage: 'Merge ' , mergeCommit id asString ].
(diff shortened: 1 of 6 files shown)


Iceberg.package/IceMerge.class/instance/updateHead.st
```diff
@@ -17,5 +17,5 @@ updateHead
 				setMergeStateBetweenCommits:
 					{self leftCommit.
 					mergeCommit}.
-			repository workingCopy commitChanges: self withMessage: 'Merge ' , mergeCommit id asString ].
+			repository workingCopy commitChanges: self withMessage: 'Merge ' , mergeCommit id asString force: true ].
```
guillep: We should put at some point a comment saying why the force here is necessary.

PR #976, opened 2018-08-14
fix-ask-everytime-conflicting-versions

### LINKED ISSUE #975: metacello asks too many times what to install when there are conflicting versions
when there is a difference between one dependency version and another (say we want to install v1 and another dependency points to v2), we need to choose. 
Due the nature of metacello, this question can be asked several times and that's very annoying. 
We can improve this situation by recording during the install session the choices made.

estebanlm:
fix #975


estebanlm:
so? I don't think the windows problems are real, but another run of the appveyor ci may be necessary.

PR #993, opened 2018-09-06
992-fix-tonel-migration

### LINKED ISSUE #992: migration tool fails to migrate a package with just extensions
Since in that case snapshot will not have a class organisation definition, it will raise a DNU.

estebanlm:
ensures a class organisation is always present when converting a package.

Fixes #992


estebanlm:
it would be good, but is complicated to reproduce. R2 packages produce always a correct package (with an MCClassOrganisation definition). 
Any idea on how to force an incorrect package?

PR #1005, opened 2018-09-27
issue1004-displayScaleFactor

### LINKED ISSUE #1004: use displayScaleFactor
The Iceberg UI should use displayScaleFactor to work better on HiDPI displays or during presentations

pavel-krivanek:
use displayScaleFactor for hardcoded dimensions
Fixes #1004


guillep:
Wow we need new icons :)

estebanlm:
> Wow we need new icons :)
in fact I'm exploring the possibility of using .icns icons instead .png (to have different sizes to choose).

estebanlm:
btw, @pavel-krivanek If you just call your branch NUMBER-name (without the issue prefix), waffle will do some movements automatically and we will all be happier :P

pavel-krivanek:
only some failing tests like: IceGenericError: curl error: Could not resolve host: github.com

PR #1027, opened 2018-10-14
Improved new tag dialog

peteruhnak:
Added buttons to create next logical semantic versions and the listing of current tags.

As there is no Semver library for pharo (afaik), the version extraction is freehanded (but I've added tests for that).

image


guillep:
Thanks!

Issue #1058, opened 2018-11-02
For Pharo 70 iceberg should not be shipped using deprecated class


Ducasse:
CmdShortcutCommandActivation -> CmdShortcutActivation

Ducasse:
https://pharo.fogbugz.com/f/cases/22612/No-DeprecatedClass-use-Iceberg-Command-use-deprecated-CmdMenuActionContext-something

Ducasse:
I issued a PR 
https://github.com/pharo-vcs/iceberg/pull/1059

guillep:
Already available in latest Iceberg

PR #1059, opened 2018-11-02
For Pharo 70 iceberg should not be shipped using deprecated class

Ducasse:
For Pharo 70 iceberg should not be shipped using deprecated class
#1058 opened 31 seconds ago by Ducasse  
https://pharo.fogbugz.com/f/cases/22612/No-DeprecatedClass-use-Iceberg-Command-use-deprecated-CmdMenuActionContext-something


jecisc:
I think deprecated classes are already removed from the development version.

Ducasse:
So I work for nothing then.

PR #1063, opened 2018-11-05
1061-Introduce-iconNamed-in-IceDefinition-and-IceTipModel-and-remove-all-the-terrible-Smalltalk-ui-icons

jecisc:
This PR remove a lot of the usage of Object>>iconNamed: 

It does not remove everything but this is a first pass.


jecisc:
Error seems not related.

PR #1075, opened 2018-11-06
847-Move-tags-under-remotes-in-Repository-view

### LINKED ISSUE #847: Move tags under remotes in Repository view
Currently, in the repository view we have:

- Local branches
- Tags
- Remotes

I think we should have 

- Locals branches
- Remotes
- Tags

In general, we have fewer remotes than tags. 

Now it's long to scroll to the remotes when we browse Pharo for example.

jecisc:
Swap remotes and tags in repository browser.

In general we have less remotes than tags for big repos.It wake it easier to work on big repos like that.

Fixes #847


jecisc:
Result:

screen shot 2018-11-06 at 15 39 07

PR #1085, opened 2018-11-08
1084-Review-design-of-outline-datasource

### LINKED ISSUE #1084: Review design of outline datasource
In the last release, I removed the outline datasource because we should be able to collapse a group of this datasource. 

In the next releases of Iceberg we should re-introduce it but with a way to collapse it. (A different way than a tree)

jecisc:
Restore outline datasource and improve the design/features.

Fixes #1084


jecisc:
I don't have my tool to do gif on OSX. You'll need to load it to see the result this time :P

PR #1101, opened 2018-12-04
1099-Gitlab-integration-should-manage-non-default-port

### LINKED ISSUE #1099: Gitlab integration should manage non default port
It would be great to be able to give a non default ssh port in a gitlab declaration.

I would like to be able to use a code like this:

```Smalltalk
Metacello new
	baseline: 'Bazard';
	repository: 'gitlab://gitlab.ferlicot.fr:3456:Projet/Bazard/src';
	load.
```

Or

```Smalltalk
Metacello new
	baseline: 'Bazard';
	repository: 'gitlab://gitlab.ferlicot.fr:3456/Projet/Bazard/src';
	load.
```

For than we need to update MCGitlabRepository class>>parseLocation:version: to extract the port then MCGitlabRepository>>scpUrl and MCGitlabRepository>>httpsUrl to use the port to get something like `ssh://git@gitlab.ferlicot.fr:3456/Projet/Bazard.git` instead of `git@gitlab.ferlicot.fr/Projet/Bazard.git`

jecisc:
Add possibility to clone project from self hosted gitlab with non default ssh port  via the Metacello integration.

Fixes #1099

/!\ This needs https://github.com/pharo-project/pharo/pull/2041 before it can work

### DIFF
--- a/Iceberg-Pharo6.package/MCGitlabRepository.extension/instance/scpUrl.st
+++ b/Iceberg-Pharo6.package/MCGitlabRepository.extension/instance/scpUrl.st
@@ -1,0 +1,6 @@
+*Iceberg-Pharo6
+sshPort
+	"In pharo 6 we do not have the code to be able to load gitlab with non default ssh port projects. So we return nil so that #scpUrl does not break."
+
+	^ nil
+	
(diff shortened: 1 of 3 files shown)


jecisc:
Can someone relaunch this build? Dependency PR was integrated in Pharo

guillep:
I've relaunched the builds

jecisc:
I’ll try to fix the PR at the end of the week probably.

gcotelli:
@jecisc try to do something similar to what I do with the original support. There's a package for Pharo 6 compatibility, you can add here the necessary extensions.

jecisc:
Updated

PR #1125, opened 2018-12-26
1124-Loading-of-Iceberg-is-broken-because-of-Commander

jecisc:
Commander is now in the src subdirectory.

Fixes #1124

PR #1127, opened 2018-12-27
Load tonel as pre-load action

guillep:
Since now commander (and possibly in the future iceberg itself and its dependencies) might be in tonel format, we need to load tonel before attempting to load. 

Follow-up of the original issue in here: https://github.com/pharo-vcs/iceberg/pull/1125


guillep:
Updating metacello seems to work for both pharo 6 and 7

image

PR #1161, opened 2019-01-30
Fix #949

### LINKED ISSUE #949: Metacello Integration fails on Missing repositories
As @LucFabresse and @Ducasse pointed out, if we download an image that was configured in a different machine (for example, built in a CI), Iceberg repositories are in a missing repository state.

In such case, the repository can be manually fixed from Iceberg's UI, but trying to execute a metacello expression to load a project that involves one of the missing repositories, it fails.

tesonep:
Fix #949 
When a repositoriy is missing and used by metacello, everything explodes. 
The repository should be recloned.


guillep:
Thanks!

PR #1173, opened 2019-02-12
1172-Iceberg-has-unused-variables

### LINKED ISSUE #1172: Iceberg has unused variables
Iceberg has classes with unused variables. I would like to remove all the unused variables of the image and then add a test so that we do not introduce new ones in Pharo.

jecisc:
Remove unused variables.

Fixes #1172


guillep:
Thanks!

PR #1176, opened 2019-02-12
1175-Iceberg-should-follow-changes-of-FastTable

### LINKED ISSUE #1175: Iceberg should follow changes of FastTable
Some methods of FastTable were deprecated in Pharo 8. 

We should add the new ones in a compatibility package for P7 and replace the deprecated calls.

jecisc:
Use new FastTable API.

Fixes #1175


Iceberg-Pharo7.package/FTTableMorph.extension/instance/deselectAll.st
```diff
@@ -0,0 +1,3 @@
+*Iceberg-Pharo7
+deselectAll
```
dionisiydk: Why it is an iceberg extension? It is usefull method and deserve to be part of fast table

Iceberg-Pharo7.package/FTTableMorph.extension/instance/selectFirst.st
```diff
@@ -0,0 +1,3 @@
+*Iceberg-Pharo7
+selectFirst
+	self selectIndex: 1
```
dionisiydk: Same here

jecisc: Those methods are in a fast table in Pharo 8. I use them in this PR and I add them in a compatibility package for Pharo 7 because they are not available in Pharo 7.

dionisiydk:
ok

чт, 14 февр. 2019 г. в 23:11, CyrilFerlicot <notifications@github.com>:

guillep:
I don't understand why this is necessary.
Can't we just make a release version of Spec and migrate iceberg to use the new spec?

estebanlm:
agree with @guillep ... better to just start migrating iceberg to the new Spec.
(... and that will be iceberg 2.0... or something like that)

jecisc:
It is currently really bothering because each time I do a contribution in Pharo 8 I need to uncheck a lot of changes that were produced by the auto deprecation.

guillep:
I'll close this one :)

PR #1178, opened 2019-02-12
1155-Disable-merge-warning

### LINKED ISSUE #1155: Disable merge warning
I was wondering if we should keep the merge warning since it is now here for some months without that many problems? 

If we keep it we should transform it into an information dialog.

jecisc:
Remove merge warning since we have it for month now without complain about it.

Fixes #1155

### DIFF
--- a/Iceberg.package/IceCommitish.class/instance/validateCanMerge.st
+++ b/Iceberg.package/IceCommitish.class/instance/validateCanMerge.st
@@ -2,9 +2,6 @@ validating
 validateCanMerge
 
 	| imageCommit headCommit mergeCommit |
-	self repository isModified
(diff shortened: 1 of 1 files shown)


guillep:
No, this change is bad. That warning is there for a reason. There is a missing feature and that's the only thing between that missing feature and a super odd behaviour.

jecisc:
Do we know what is missing? Because the warning does not says there is a missing feature, just that it is experimental and I do not remember someone complaining because the merge failed (but I might just have missed it)

guillep:
It's there in the warning message:

`merge when there is a dirty working copy. Could cause a loss of your local changes`

PR #1189, opened 2019-02-20
added documentation for hot fixes

SabineMa:
please correct if something is wrong


guillep:
This is super great :)!!!
Let's integrate it and we can iterate it later. 
Thanks a lot @SabineMa !

PR #1191, opened 2019-02-21
Migrate iceberg sources to Tonel

guillep:
This should work with latest changes in Tonel dependencies and installation scripts


guillep:
I'll remake this PR

PR #1192, opened 2019-02-21
Migrate sources to tonel

guillep:
Use tonel format:
 - more compact
 - more readable

The newer changes in the tonel dependencies should allow us to do it without much effort now.


guillep:
Test failures not related (broken from before).
Still migration went ok: could load code and run tests

PR #1202, opened 2019-02-25
update tutorial link in README

hogoww:
You should still check if the link is the resource that you wanted to link to, in case i misunderstood. 
The previous link is dead anyway.
Pierre.

### DIFF
--- a/README.md
+++ b/README.md
@@ -150,4 +150,4 @@ Metacello new
 
 ## 5 minutes tutorial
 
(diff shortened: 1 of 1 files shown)


hogoww:
I'm pretty sure the right link should be "https://github.com/pharo-vcs/iceberg/wiki/Tutorial", but i'm not sure given the name of file in the initial link.

guillep:
I'd prefer the second link as you say, could you update the PR? :)

hogoww:
Done !

PR #1212, opened 2019-03-08
1211-Some-presenters-are-overriding-extent-instead-of-initialExtent

### LINKED ISSUE #1211: Some presenters are overriding #extent instead of #initialExtent
This is not the right way and will probably lead to glitches with the updated Spec.

jecisc:
Do not override ComposablePresenter>>extent

Fixes #1211


guillep (approved):
Ok!

PR #1213, opened 2019-03-13
Fix migration of tagged packages

estebanlm:
Fixes pharo-vcs/tonel#74


estebanlm:
based in pre-tonel version, I will submit a new onw

Issue #1217, opened 2019-03-20
Introduce Hiedra to repository browser


tinchodias:
The latest version of Hiedra should be good in performance.

PR #1218, opened 2019-03-20
1217-Introduce-Hiedra-to-repository-browser

tinchodias:
Add a Hiedra column to the Repository window. 

I adds a delay when the window is opened because it is not asynchronous. But I think the information it adds makes it worth. 
In small branches it was a few milliseconds in my laptop, and the worst case I tested is 500ms slowdown with the dev-1.0 branch of the iceberg repository (>2800 commits).
The rendering part is not significant in performance because it uses a kind of pagination. For example, if a fasttable is opened and only the top 20 rows are visible, only one "page" of 30 rows are rendered and cached initially. When user scrolls the table, more pages of 30 rows are rendered on demand.
What could be optimized yet is the graph model: when the fasttable is opened, Hiedra currently iterates over all the commits and parents to create the model. But I don't plan to work in this optimization for the moment... so please test it.

IMPORTANT: It works in latest Pharo 8 + PR #2926 (https://github.com/pharo-project/pharo/pull/2926).

### DIFF
--- a/Iceberg-TipUI/IceTipHiedraHistoryBrowser.class.st
+++ b/Iceberg-TipUI/IceTipHiedraHistoryBrowser.class.st
@@ -1,0 +1,28 @@
+"
+I add a Hiedra column to the browser defined by my superclass. This column visualizes the relationship between commits in the history list.
+"
+Class {
+	#name : #IceTipHiedraHistoryBrowser,
+	#superclass : #IceTipHistoryBrowser,
+	#category : #'Iceberg-TipUI-View'
+}
+
+{ #category : #initialization }
+IceTipHiedraHistoryBrowser >> initializeCommitList [
+
+	"Add the Hiedra column at the beginning. The Datasource has the logic to render it."
+	commitList widget addColumn:
+		(IceTipTableColumn new
+			id: ' ';
+			yourself).
+
+	super initializeCommitList.
+]
+
+{ #category : #'private factory' }
+IceTipHiedraHistoryBrowser >> newCommitsDataSource [
+	^ IceTipHiedraDataSource new 
+		tool: self;
+		elements: #(); "It will be set when refreshing"
+		yourself
+]
--- a/Iceberg-TipUI/IceTipRepositoryBrowser.class.st
+++ b/Iceberg-TipUI/IceTipRepositoryBrowser.class.st
@@ -80,7 +80,7 @@ IceTipRepositoryBrowser >> initializeWidgets [
 	super initializeWidgets.	
 	
 	sidebarTree := self newIceOutline.
-	historyPanel := self instantiate: IceTipHistoryBrowser on: self model headModel.
+	historyPanel := self instantiate: IceTipHiedraHistoryBrowser on: self model headModel.
 	historyPanel beForMerge.
 	
 	self initializeSidebarTree.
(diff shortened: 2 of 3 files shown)


tinchodias:
screenshot

tinchodias:
Tested also in Dark Theme.

guillep:
Fixes #1217

tinchodias:
Ohhhh, and what can we do with incompatibilities in Pharo 7?

guillep:
Ahh, I thought that was already there.
@tinchodias what do you propose? I see two solutions, maybe there is something else to take into account:
 - We could make Iceberg's hiedra independent of Epicea's one, and have an entire copy of hiedra for iceberg. This would work for pharo7 and 8.
 - we could just hide/disable the hiedra for Pharo7. This should be easy to implement

tinchodias:
I propose 2nd option

Iceberg-TipUI/IceTipHiedraHistoryBrowser.class.st
```diff
@@ -0,0 +1,28 @@
...
+}
+
+{ #category : #initialization }
+IceTipHiedraHistoryBrowser >> initializeCommitList [
+
+	"Add the Hiedra column at the beginning. The Datasource has the logic to render it."
+	commitList widget addColumn:
+		(IceTipTableColumn new
+			id: ' ';
+			yourself).
```
guillep: @tinchodias Would it be enough to make this line to execute only for Pharo8?

Iceberg-TipUI/IceTipRepositoryBrowser.class.st
```diff
@@ -80,7 +80,7 @@ IceTipRepositoryBrowser >> initializeWidgets [
 	super initializeWidgets.	
 	
 	sidebarTree := self newIceOutline.
-	historyPanel := self instantiate: IceTipHistoryBrowser on: self model headModel.
+	historyPanel := self instantiate: IceTipHiedraHistoryBrowser on: self model headModel.
```
guillep: @tinchodias Or we should make it conditional here to use IceTipHistoryBrowser or IceTipHiedraHistoryBrowser?

tinchodias: It could use a method like this, what's your opinion?
~~~
historyBrowserClass
	"Answer the Presenter class to instantiate for browsing the history of the repository.
	The choices are IceTipHistoryBrowser or IceTipHiedraHistoryBrowser (the case if Hiedra v2.0.0 is loaded)."

	^ (self class environment hasClassNamed: #HiColumnController)
		ifTrue: [ IceTipHiedraHistoryBrowser ]
		ifFalse: [ IceTipHistoryBrowser ]
~~~

Iceberg-TipUI/IceTipHiedraHistoryBrowser.class.st
tinchodias: The other option is better

Iceberg-TipUI/IceTipRepositoryBrowser.class.st
tinchodias: I mean, then the line should be:
~~~
historyPanel := self instantiate: self historyBrowserClass on: self model headModel.
~~~

tinchodias: Note: I pressed on "start review" by mistake.

PR #1219, opened 2019-03-23
feature/add-otp-support

zecke:
Initial and minimal support for OTP based 2factor authentication for github.

two (2) factor authentication is an extra security measure for a github account and requires providing a one time token as part of the authentication.  Add a minimal implementation (the bigger one requires oauth to create a long lived token). Add a subclass for the credentials that can store the tokens and then use it when approriate. Do double dispatch on it. Avoid slicing back to the default plain text type.

Details:
https://developer.github.com/v3/auth/#working-with-two-factor-authentication


zecke:
I rebased this from the pharo8 version to master. The PR was created with the patch.

Iceberg-Plugin-GitHub.package/IceGitHubAPI.class/instance/newRequestTo..st
```diff
@@ -2,10 +2,8 @@ private factory
 newRequestTo: aString
 	^ ZnClient new 
 		in: [ :this | 
-			isAnonymous ifFalse: [ 
-				this
-					username: self credentials username 
-					password: self credentials password ] ]; 
+			isAnonymous ifFalse: [
+				self credentials applyToRequest: this ] ];
```
zecke: Something strange happened in a fresh image. "self credentials" was actually an instance of IceTipInteractiveCredentialsProvider. I am not sure how this happened. It would have DNUed on username as well.

guillep:
Hi Holger, the main development branch is `dev-1.0` and it has moved to tonel some time ago already... Could you redo the changes on that branch? Otherwise I can try to do it myself in some days but not now...

guillep:
I've merged it manually because this change was pre migration to tonel.
To my surprise, iceberg manages well merges between tonel and filetree :).
This will be in the next release

PR #1220, opened 2019-03-23
Update Commander version 0.8.0.

dionisiydk:
Update Commander version 0.8.0.
It is simple cleanup of isAbstract implementation. 
It does not affect Iceberg because it initialy uses right approach (base Iceberg command defines isAbstract with "^self = MyClass")


guillep:
I'll redo it myself, Iceberg has been migrated to tonel.

PR #1232, opened 2019-04-29
Small improvement for user understanding for error when cloning a pri…

hogoww:
…vate repository.

That way, at least people won't be as surprised with this error, even though it's not a fix :/
Feel free to be more precise, or correct me if I was mistaken !

I only wanted to add the new comment in #responseWithValidationDo:, but despite having taken the last repository, some meta data were commited aswell. Sorry.

### DIFF
--- a/Iceberg-Plugin-GitHub.package/IceGitHubAPI.class/instance/responseWithValidationDo..st
+++ b/Iceberg-Plugin-GitHub.package/IceGitHubAPI.class/instance/responseWithValidationDo..st
@@ -7,5 +7,11 @@ responseWithValidationDo: aBlock
 	self handleUnauthorized: response.
 	
 	response isSuccess 
-		ifFalse: [ IceGitHubError signalFor: response ].
+		ifFalse: [ 
+			"Your repository is already cloned, but without metadata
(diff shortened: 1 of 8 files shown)


Iceberg-Plugin-GitHub.package/IceGitHubAPI.class/instance/responseWithValidationDo..st
```diff
@@ -7,5 +7,11 @@ responseWithValidationDo: aBlock
 	self handleUnauthorized: response.
 	
 	response isSuccess 
-		ifFalse: [ IceGitHubError signalFor: response ].
+		ifFalse: [ 
+			"Your repository is already cloned, but without metadata
+	Small fix is to 'import from existing clone' to have your repository displayed in Iceberg.
+	
+	This appears because Iceberg is trying to get the repository meta data without having your credentials which isn't possible on a private repository."
+			IceGitHubError signalFor: response
```
guillep: I don't think normal users will really benefit from a comment...

guillep:
I'll close this PR. At best this should be done starting from the dev branch

PR #1245, opened 2019-05-30
Adapt UI so the text is not cut on the right

SergeStinckwich:
The text was crop on the right.

Screenshot 2019-05-30 at 15 27 06


SergeStinckwich:
What happens here ?

SergeStinckwich:
What is the situation on this one ? why not integrate ?

guillep:
I did the merge manually because this change predates the tonel migration, and git nor github know how to merge tonel and filetree, but iceberg does :)

This will be in the next release

tinchodias:
Support merge between branches with different format is really outstanding.

PR #1250, opened 2019-06-17
1249-Remove-calls-to-deprecated-methods-

### LINKED ISSUE #1249: Remove calls to deprecated methods
In Spec 2, all methods related to window management were deprecated in ComposablePresenter. 

Iceberg should be updated to follow those changes.

jecisc:
Remove use of some deprecated methods of Spec 2 and FastTable.

WARNING: This breaks the P7 compatibility, but after a talk with Pablo and Esteban we think it's probably good because Iceberg from P7 is already pretty stable.

Fixes #1249


jecisc:
How can I have conflict when I started my changes from dev-1.0 after doing a pull?!

jecisc:
I pulled the wrong remote :(

jecisc:
I pulled the wrong remote :(

jecisc:
I pulled the wrong remote :(

PR #1261, opened 2019-06-27
Add #httpUrl as a valid option for #remoteTypeSelector.

### LINKED ISSUE #1260: HTTP remote type
Sometimes you want to access a remote using plain HTTP and not HTTPS, for example for self-hosted instances. 
In that case you need something like a HTTP remote type.

gcotelli:
Also categorize some uncategorized methods.

Fixes #1260


gcotelli:
CI failures are totally unrelated. Seems to be a general problem with the dev branch

guillep:
I merged it upstream, it'll be there for the next release. Thanks!

PR #1264, opened 2019-06-28
1263 - Do not use default name in toolbar items but real command name

### LINKED ISSUE #1263: IceTipToolbarActivation label should not be the default command name

jecisc:
The goal is to be able to personalize the name depending on where is the command.

Fixes #1263


jecisc:
Oh, I see that there is "defaultToolbarItemName" already. So this is not needed.

PR #1267, opened 2019-07-03
1266-Iceberg-should-not-use-Spec-internals

### LINKED ISSUE #1266: Iceberg should not use Spec internals
Spec directly use a value holder of Spec instead of using Spec public API. This will break with Spec 2.

jecisc:
Remove call to Spec internals.

Fixes #1266

### DIFF
--- a/Iceberg-Plugin-GitHub/IceGitHubNewBranchFromIssuePanel.class.st
+++ b/Iceberg-Plugin-GitHub/IceGitHubNewBranchFromIssuePanel.class.st
@@ -109,8 +109,7 @@ IceGitHubNewBranchFromIssuePanel >> initializeWidgetsContents [
 		whenBuiltDo: [ :w | 
 			w widget wrapFlag: false.
 			w widget enabled: false ].
-	issueNumberText textHolder
-		whenChangedDo: [ :text | self validateIssue: text ]
+	issueNumberText whenTextChangedDo: [ :text | self validateIssue: text ]
 ]
 
 { #category : #'accessing ui' }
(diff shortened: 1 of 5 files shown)


gcotelli (approved):

Iceberg-Plugin-Pharo/IcePharoNewBranchFromIssuePanel.class.st
```diff
@@ -96,7 +96,7 @@ IcePharoNewBranchFromIssuePanel >> initializeWidgetsContents [
 		whenBuiltDo: [ :w | 
 			w widget wrapFlag: false.
 			w widget enabled: false ].
-	issueNumberText textHolder whenChangedDo: [ :text | self validateIssue: text ]
+	issueNumberText whenTextChangedDo: [ :text | self validateIssue: text ]
```
tinchodias, line 99: In dev-1.9, this change is already applied

Iceberg-Plugin-GitHub/IceGitHubNewBranchFromIssuePanel.class.st
```diff
@@ -109,8 +109,7 @@ IceGitHubNewBranchFromIssuePanel >> initializeWidgetsContents [
 		whenBuiltDo: [ :w | 
 			w widget wrapFlag: false.
 			w widget enabled: false ].
-	issueNumberText textHolder
-		whenChangedDo: [ :text | self validateIssue: text ]
+	issueNumberText whenTextChangedDo: [ :text | self validateIssue: text ]
```
tinchodias, line 112: In dev-1.9, this change is already applied

tinchodias:
Should we close?

tinchodias:
Thanks for the changes, they are in dev-1.9.

PR #1269, opened 2019-07-16
Migrating iceberg windows from spec1 to spec2

Inao0:


Iceberg-Plugin-Pharo/IcePharoNewBranchFromIssuePanel.class.st
```diff
@@ -84,19 +84,19 @@ IcePharoNewBranchFromIssuePanel >> initializeWidgetsContents [
...
+		placeholder: 'e.g., a-fogbugz-issue';
 		autoAccept: true.
-		
 	issueNumberText whenBuiltDo: [ :w | w widget wrapFlag: false ].
 	issueText
 		whenBuiltDo: [ :w | 
 			w widget wrapFlag: false.
 			w widget enabled: false ].
-	issueNumberText textHolder whenChangedDo: [ :text | self validateIssue: text ]
+	issueNumberText textHolder
```
jecisc: You should not use textHolder but whenTextChangedDo:

Iceberg-Plugin-GitHub.package/IceGitHubCreatePullRequestModel.class/instance/initializeWidgets.st
```diff
@@ -0,0 +1,26 @@
...
+	headLabel := self newLabel.
+	headForkList := self newDropList.
+	headBranchList := self newDropList.
+	baseLabel := self newLabel.
+	baseForkList := self newDropList.
+	baseBranchList := self newDropList.
+	bodyLabel := self newLabel.
+	bodyPanel := self newText autoAccept: true.
+		
+	titleLabel label: 'Title'.
```
jecisc: Static labels do not need a LabelPresenter anymore. You can just add a string in the layout.

tesonep: Why you can put strings in a layout? so the string is polymorphic with a presenter?

jecisc: If it is a String and not a Symbol representing the name of a variable containing a presenter, it will automatically consider it as a LabelPresenter.

guillep: I think what pablo means is not "how does it work", but the fact that strings and presenters are polymorphic is confusing and potentially dangerous.

jecisc: What is currently done is that if the argument is neither a symbol or a layout, it will call #asPresenter on it. 

On strings, it will return a LabelPresenter. 

For now, most people who learned you can do that in Spec 2 seemed not confused and pretty happy with that because it simplifies a lot of presenters.

guillep: I understand how it works, but thanks :).

I do not agree that it simplifies a lot (a couple of lines?).
And I do not agree that Iceberg needs that style of coding.

Iceberg-Plugin-Pharo/IcePharoNewBranchFromIssuePanel.class.st
guillep: Shouldn't these holders be deprecated and rewritten in spec with spec observable slots?

jecisc: Those accessors to holders are already deprecated in Spec2 since classes using presenters should not use the internals of the presenters when the API already provides a way to do what the users want. 

There is not an automatic rewrite because what should be replaced will depend on what the user did with the holder.

guillep: ok! I wonder why the did not catch the deprecation then...

Iceberg-Plugin-GitHub.package/IceGitHubCreatePullRequestModel.class/instance/initializeWidgets.st
estebanlm: Well, I added it :)
And I think Is a big simplification for coding form-style layouts, which are a lot of the dialogs you normally do. 
I do not understand why you need to do the double work of defining a presenter and assigning a label when you can do it directly in the layout avoiding the overhead.

Issue #1221, opened 2019-03-28
History browser install action does not work on the class side


jecisc:
I changed a method in a class and wanted to revert it. So I opened the History Browser and tried to install the latest committed version of the method. The problem is that I was in class side and the installed version was compiled on the instance side.

guillep:
This seems to be a bug between the `IceLibgit**LogReader` and the conversion to ring definitions.

```smalltalk
(IceLibgitFiletreeLogReader fileName: aPath on: aCommit)
		package: aMethod package mcPackage; 
		packageDirectory: (self repository directoryPathStringForPackage: aMethod package);
		definitions)
```

The ring definition does not realize the method was actually from class side, because it checks only the full name of the class, and seems to expect a name like `ABlah class`.

The other related method is:

```smalltalk

collectVersionsFrom: aCollection method: aMethod path: aPath
	| lastDefinition basepath history |
		
	basepath := self basepathOf: aPath.
	lastDefinition := MCMethodDefinition new.
	history := OrderedCollection new.
	aCollection reverseDo: [ :eachCommit | | definition |
		definition := self definitionFor: aMethod path: basepath commit: eachCommit.
		lastDefinition = definition  ifFalse: [ 
			1halt.
			history add: (IceLogVersion 
				commit: (self repository commitFromGitCommit: eachCommit)
				definition: definition asRingDefinition).
			lastDefinition := definition ] ].

	^ history reversed
```

ironirc:
This changed method solves it.
```
IceTipVersionModel>>#basicInstall
    | definition cls |
    definition := self entity definition.
    cls := definition realClass.
    definition isMetaSide ifTrue: [ cls := cls class ].
    cls compile: definition sourceCode classified: definition category
```

I find it a bit strange that RGMethodDefinition shows the method in the form:
`Hobby>>#current` instead of `Hobby class>>#current`, despite isMetaSide being true.
I would also expect that `definition realClass` already would render the metaClass. Probably leading us into rabbit holes ;-)

Ducasse:
Tx :)

jecisc:
@AlexisCnockaert can you check if this issue is still valid? If yes can you try to fix it wit what was said in the discussion?

The history browser can be open via this icon:
<img width="24" height="29" alt="Image" src="https://github.com/user-attachments/assets/bc537ef8-7e0c-4ee8-8758-9453b5e8bba6" />

In calypso

[...]

PR #1278, opened 2019-07-19
Likely fix for issue #1221

dahoward:
We seem to have found the cause of this bug after debugging the command behind the morph toolbar button.
We expect this change to fix issue #1221.


guillep:
I just tried this change and it does not fix the issue.
Nevertheless, the issue does still exist.

PR #1283, opened 2019-07-30
Add a smart cherry picking window

Inao0:


estebanlm:
what is a "smart cherry picking window" ?

Issue #1284, opened 2019-07-30
Commit window should have a context menu in the tree.


tesonep:
![image](https://user-images.githubusercontent.com/355455/62126479-064a8700-b2d0-11e9-9e7f-881ff970e760.png)

Inao0:
Fixed by pull request #1285

PR #1285, opened 2019-07-31
Allows to automatically save and push from commit browser

Inao0:
The values are stored in Iceberg settings.


Inao0:
Should Fix #1284

PR #1321, opened 2019-10-21
remove-dependency-to-glamour

estebanlm:
remove dependency of glamour


estebanlm:
this is seriously wrong

PR #1323, opened 2019-10-21
remove-glamour-dependency-2

estebanlm:
fix announcer to not depend on GLMAnnoucner

see if this works?

### DIFF
--- a/Iceberg/IceMCVersionInfo.class.st
+++ b/Iceberg/IceMCVersionInfo.class.st
@@ -18,7 +18,7 @@ IceMCVersionInfo class >> package: package message: commitMessage [
 ]
 
 { #category : #utilities }
-IceMCVersionInfo class >> uuidFromCommit: aCommit package: aPackage. [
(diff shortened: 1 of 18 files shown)


Iceberg/IceMCVersionInfo.class.st
```diff
@@ -18,7 +18,7 @@ IceMCVersionInfo class >> package: package message: commitMessage [
 ]
 
 { #category : #utilities }
-IceMCVersionInfo class >> uuidFromCommit: aCommit package: aPackage. [
+IceMCVersionInfo class >> uuidFromCommit: aCommit package: aPackage [.
```
jecisc:
```suggestion
IceMCVersionInfo class >> uuidFromCommit: aCommit package: aPackage [
```

jecisc:
There is a dot that should not be here I think, except that, the change seems fine. :)

estebanlm:
where?

jecisc:
I did a suggestion of code change just above

PR #1341, opened 2020-01-27
5578-Iceberg-method-history-shows-only-oldest-commit

akgrant43:
IceLibgitFiletreeLogReader>>loadDefinitions set the entryPath relative to the git repository root, not just the package name.

Fixes: https://github.com/pharo-project/pharo/issues/5578


guillep:
Thanks!

PR #1358, opened 2020-04-01
1356-DNU-when-pasting-url-of-github-project-to-Clone-remote-Repository-Regression

tesonep:
The text changed announcement can come with a Text, so we need to convert it to String if needed.

This is a problem of the under deprecation Spec1, I fix it here because modifying Spec1 to handle both cases is more work and it is something we are going to throw away.


guillep:
Fixed upstream

PR #1360, opened 2020-04-06
made the caches threadsafe

georgeganea:
The views in gt are rendered in a bg thread and because of that they can access the same repository instance at the same time. This can lead to errors like the one in the attached screenshot. 
Screenshot 2020-04-01 at 16 53 57

### DIFF
--- a/Iceberg-Libgit/IceLibgitRepository.class.st
+++ b/Iceberg-Libgit/IceLibgitRepository.class.st
@@ -303,7 +303,7 @@ IceLibgitRepository >> codeDirectory [
 { #category : #'private-commits' }
 IceLibgitRepository >> commitCache [
 	
-	^ commitCache ifNil: [ commitCache := LRUCache new maximumWeight: 30  ]
+	^ commitCache ifNil: [ commitCache := LRUCache new maximumWeight: 30; beThreadSafe  ]
 ]
 
 { #category : #'private-commits' }
--- a/Iceberg/IceRepository.class.st
+++ b/Iceberg/IceRepository.class.st
@@ -151,7 +151,7 @@ Class {
 	#classVars : [
 		'Registry'
 	],
-	#category : 'Iceberg-Core'
+	#category : #'Iceberg-Core'
 }
 
 { #category : #testing }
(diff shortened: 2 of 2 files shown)


tesonep:
I don't like this solution, basically, I think about the following problems:

- what are the causes of the race condition? Is this solution handling it, we have not a test to see the problem. 

- Making it threadsafe like this is just penalizing all the users, even the ones are not producing the race condition. How we can measure the impact of this change. Has it been measured? For example, loading big projects use heavily such code, so they will be impacted by this change.

- To me, this is a problem that should be solved in the users of it. The users of it in the Iceberg UI already caches some of the values, we cannot expect that accessing the inner layers of the tool will work in all the scenarios. It was never intended so, that is why we are not documenting it extensible. 

- Also, this is not good, it will explode later, as libgit is not threadsafe.  It's threadsafe if you are not sharing the repositories. So performing different operations on libgit on the same repository can have unknown results, even producing wrong commits and invalid state.

To me this is not a bugfix Iceberg does not have this problem, if it is used in a way it was not intended we cannot guarantee anything. It is a new feature and should be heavily analyzed as to what it is. Even if at the end the solution is a single line, we need to be sure the impact of that line.

guillep:
I kind of agree with Pablo, could you give more details? Because this **fix** smells like a patch. Iceberg is not thread safe, and just changing a cache will maybe be a fix for the immediate case you have, but you'll find other problems in the future ("pan para hoy, hambre para mañana" we say in spanish => bread for today, hunger for tomorrow).

I mean, you may find later that the `IceRepository` should be treadsafe too, so you may want to add a monitor and to protect all methods with a critical section?

Hope you can share more with us.

georgeganea:
First of all, sorry for replying so late. I somehow missed your replies... 

The usecase is a visualization that displays all commits in a tree and naturally that takes some time to compute for a large repo. At the same time, the user can be viewing the same repo via another view and since we are displaying the same instance of IceLibgitRepository, the cache error would show up.

The solution will be that we will use a different instance of IceLibgitRepository for each gtView. This does not imply any change in Iceberg. I will close this pull request.

Thank you for your input!

PR #1370, opened 2020-05-11
Fix typo

crunsk:
Changed 'atttached' to 'attached'.

### DIFF
--- a/Iceberg-TipUI/IceTipRepositoryModel.class.st
+++ b/Iceberg-TipUI/IceTipRepositoryModel.class.st
@@ -88,7 +88,7 @@ IceTipRepositoryModel >> branches [
 IceTipRepositoryModel >> checkMissing [
 	
 	self isMissing ifTrue: [ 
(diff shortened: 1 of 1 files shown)


astares:
Note this is against "dev-1.0" branch instead of "dev-1.6".

crunsk:
> 
> 
> Note this PR is against "dev-1.0" branch instead of "dev-1.6".

Should I do this change against the branch  "dev-1.6" and then do a pull request?

guillep:
> Should I do this change against the branch "dev-1.6" and then do a pull request?

I've just tried to make the pull request target branch dev-1.6, but it came with lots of noise... I don't think github is smart enough to make this merge...

So, yes, could you do it against 1.6? It could be merged right away.

Thanks BTW!

crunsk:
Ok, [here's a link](https://github.com/pharo-vcs/iceberg/pull/1371) to the new pull request. Hopefully that's ok. :)

guillep:
We can close this one now.

Thanks again!

PR #1373, opened 2020-05-27
Add git config instructions

LTangaF:
Add some documentation on how to set config items from within Pharo


guillep:
I'll merge it in dev-8 which is now the default branch!

Thanks!

PR #1381, opened 2020-07-17
missing polymorphism

demarey:


tinchodias:
Only windows CI failed with unrelated errors (I think)

PR #1399, opened 2021-01-05
add subprojects to Pharo plugin

estebanlm:
the idea is to have 

pharo
pharo-spec2
pharo-newtools 

as predefined projects in pharo.
Maybe we can add also pharo-iceberg?

Maybe I need to do this directly to dev-1.9?


tinchodias:
I merged it without testing, but trust you, and also was a good reason to try the "PR browser" (that was recently migrated in dev-1.9).

PR #1412, opened 2021-02-11
Move Iceberg menu item from Library to Sources.

gcorriga:
Following the recent reorg of the world menu in Pharo 9 we need to move the Iceberg menu item from its current location in the Library menu to the Sources menu.


tinchodias:
thanks! we will try it in dev-2.0 and merge into pharo 9 with next Iceberg release

PR #1439, opened 2021-03-01
Keep difference between open dialog and open modal

### LINKED ISSUE #1438: Fix tests in IceTipAskForPlaintextCredentialsModelTest
I think this portion of the stack trace shows the problem:

```
IceTipAskForPlaintextCredentialsModel(SpAbstractPresenter)>>openModalWithSpec
IceTipAskForPlaintextCredentialsModel(IceTipDialog)>>openDialogWithSpec
IceTipAskForPlaintextCredentialsModel(IceTipAbstractCredentialsModel)>>openNonModal
IceTipAskForPlaintextCredentialsModelTest>>testAcceptCreatesCredential
```

The test executes `openNonModal` to fill some fields afterwards and then do `triggerOkAction`. But now it's stuck on the first step.

tinchodias:
Intends to fixes #1438

### DIFF
--- a/Iceberg-TipUI/IceTipDialog.class.st
+++ b/Iceberg-TipUI/IceTipDialog.class.st
@@ -29,10 +29,3 @@ IceTipDialog >> initializeDialogWindow: aWindow [
 		  okAction: [ self accept ];
 		  cancelAction: [ self cancel ].
 ]
-
-{ #category : #showing }
-IceTipDialog >> openDialogWithSpec [
-
-	^ self openModalWithSpec
-
-]


tesonep:
@tinchodias we cannot do this. Because you are breaking the behavior of callbacks in general, and in particular the one asking credentials. All Dialogs are expected to be modal in this case, if not they should do something about it. 

If you do this, you will need to see all users of dialogs to see if it has to use the modar or non modal version

tesonep (changes requested):
No

tinchodias:
Ah, now I got it.
At least the test, uses openNonModal.

PR #1445, opened 2021-03-08
1444-Repository-Window-is-loosing-the-toolbar-when-selecting-rows

### LINKED ISSUE #1444: Repository Window is loosing the toolbar when selecting rows

tesonep:
RepositoryBrowser should update the Toolbar when there is a selection change

Fix #1444


tinchodias (approved):
I couldn't reproduce the bug, but I trust you!

PR #1446, opened 2021-03-08
1396-Discard-All-button-should-ask-for-confirmation-first

### LINKED ISSUE #1396: Discard All button should ask for confirmation first
The "Discard all" button is close to the "Add package" button and does not ask the user if the changes should be really discarded.

Today after modifying some packages and adding new packages to the image I wanted to add the new packages to my Iceberg project. As the buttons are close together I accidentially hit the  "Discard all" instead of the "Add package" button.

This discarded many changes I applied on the modified packages that were already in the project. So I lost code with this.

=> we should build in a dialog "Do you really want to discard your changes" with YES/NO that the user needs to confirm first.

tesonep:
Adding a new dialog to show the changes that are going to be impacted by discarding all changes.

Fix #1396

### DIFF
--- a/Iceberg-TipUI/IceTipPreviewBrowser.class.st
+++ b/Iceberg-TipUI/IceTipPreviewBrowser.class.st
@@ -82,5 +82,11 @@ IceTipPreviewBrowser >> mainList [
 
 { #category : #'private factory' }
 IceTipPreviewBrowser >> newActionButton [
-	self subclassResponsibility
+
+	^ self newButton
+		label: 'Discard All';
+		icon: (self iconNamed: #refresh);
+		help: ('Discard all changes from the image and loads the current commit');
+		action: [ self doDiscardAll ];
+		yourself
 ]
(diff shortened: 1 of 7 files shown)


Iceberg-TipUI/IceTipPreviewBrowser.class.st
```diff
@@ -82,5 +82,11 @@ IceTipPreviewBrowser >> mainList [
...
 IceTipPreviewBrowser >> newActionButton [
-	self subclassResponsibility
+
+	^ self newButton
+		label: 'Discard All';
+		icon: (self iconNamed: #refresh);
+		help: ('Discard all changes from the image and loads the current commit');
+		action: [ self doDiscardAll ];
+		yourself
 ]
```
tinchodias: This must be in the subclass, IceTipDiscardAllChangesPreviewBrowser

tinchodias:
Wonderful! I love it.

Only found this glitch with layout when changing the size of the window with my mouse cursor:

Screen Shot 2021-03-09 at 13 40 24

And the method to push to the subclass.

tinchodias:
I opened #1452 to tackle the hidden button problem not only in this Discard all window, but also in Preview and other windows.

What do you prefer, @tesonep , about the method to push down to subclass? just merge like this and then fix in another change, or fix it before merge?

tinchodias:
A comment about a detail: the window's title. I'd prefer something shorter like "Preview Changes to Discard". Or at least do not say "sync" or "current commit".

tinchodias (approved):

PR #1480, opened 2021-04-20
make sure the first line of a method does not contain a .

### LINKED ISSUE #1477: Cleanup Periods in message signature
These methods have a . in the first line:

ManifestIcebergPluginGitHub class>>#description
ManifestIcebergPlugin class>>#description
ManifestIcebergMetacelloIntegration class>>#description
ManifestIcebergLibgitTonel class>>#description
ManifestIcebergLibgitFiletree class>>#description
ManifestIcebergLibgit class>>#description
ManifestIceberg class>>#description
IcePluginManager>>#initializeRepository:
IceMCVersionInfo class>>#uuidFromCommit:package:

MarcusDenker:
fixes #1477

### DIFF
--- a/Iceberg-Libgit-Filetree/ManifestIcebergLibgitFiletree.class.st
+++ b/Iceberg-Libgit-Filetree/ManifestIcebergLibgitFiletree.class.st
@@ -8,7 +8,9 @@ Class {
 }
 
 { #category : #'meta-data' }
-ManifestIcebergLibgitFiletree class >> description [ ^ 'I contain the libgit binding for the FileTree export format. 
+ManifestIcebergLibgitFiletree class >> description [
+
+	^ 'I contain the libgit binding for the FileTree export format. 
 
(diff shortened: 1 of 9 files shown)


astares (approved):

Iceberg-Libgit-Filetree/ManifestIcebergLibgitFiletree.class.st
```diff
@@ -8,7 +8,9 @@ Class {
 }
 
 { #category : #'meta-data' }
-ManifestIcebergLibgitFiletree class >> description [ ^ 'I contain the libgit binding for the FileTree export format. 
+ManifestIcebergLibgitFiletree class >> description [
```
tinchodias, line 11: There isn't a . in first line here, no?

tinchodias (approved):

Iceberg-Libgit-Filetree/ManifestIcebergLibgitFiletree.class.st
MarcusDenker: the method is just one line.. it was auto-generated. The test I did really just works on the string level "give me the first line, check if there is a .". We should improve that and do it on the AST level... but I think just fixing those methods is not wrong: we should not have methods without a newline after the signature in the system, either.

tinchodias: +1 thanks

PR #1501, opened 2021-06-07
[Bug]: Ice Method Should Use #contextClass for Browsing

seandenigris:
- `#canBeBrowsed` wasn't taking into account whether we have a metaclass
- `browse` was duplicating built-in behavior


seandenigris:
Errors look unrelated. @tesonep Anyone specific I should tag?

PR #1522, opened 2021-07-12
1521-Keyboard-arrows-does-not-work-in-commit-window-after-selecting-tree

tesonep:
Commands should not have a shortcut for the arrows.
Shortcuts in commands are general to the window an they do not respect the focus of the selected widget.
Arrows should be handled with bindings to the presenter.
Those bindings respect the focus.


tesonep:
Fix #1521

PR #1539, opened 2021-09-24
10015-fixing-removal-of-classInstVars

tesonep:
If the new class definition is empty it is because it does not have class instance variables.

As we cannot remove the existing definition, we need to replace it with one without classInstanceVariables.
We modify the patcher to have the new definition without class inst variables.


tesonep:
Fix https://github.com/pharo-project/pharo/issues/10015

PR #1548, opened 2021-10-22
CI: use Pharo 10

tinchodias:
I suppose it's good to move. I kept P9 in ubuntu+macos... it should be enough. For windows, there is a special consideration in the yml file so P9 is excluded.


demarey:
outdated. replaced by #1565

PR #1563, opened 2022-01-21
commit auto-rewrites from spec

demarey:


MarcusDenker (approved):

tinchodias:
I like the change. Where is this API currently available? in P10 I assume

tinchodias:
Ah, I just read the email thread with the context!

demarey:
now outdated since #1566 integration

PR #1566, opened 2022-01-25
adapt Iceberg to Spec2 changes

demarey:


tesonep:
@demarey pay attention that there are some overridden implementations of openWithSpec, openDialogWithSpec and openModalWithSpec. We need to rename also the overridden version in the subclasses.

demarey:
@tesonep I think all implementors / senders have been renamed

tesonep:
Yes, it is true. I will merge it

PR #1568, opened 2022-01-26
add command to toggle lock/unlock

estebanlm:
this just add a command to use the lock/unlock capability (now working) of metacello :

image


estebanlm:
and now we have also nice icon indicators: 

image

(this are icons already present in the image, do not ask for much more fancy stuff ;) )

PR #1583, opened 2022-04-01
fix layouts according to Spec2

demarey:


estebanlm:
sorry, I went other path before seeing this

PR #1609, opened 2022-09-01
Treat trait class side definitions like class class side definitions too

### LINKED ISSUE #1608: Adding a used trait to a trait with class side method raises an error in Pharo10
1. Take a brand new Pharo 10 image
2. Create two traits 
Trait << #TMyTest1
	package: 'MyPackage1-Exception'
 and 
Trait << #TMyTest2
	package: 'MyPackage-Exception'
in TMyTest2 add a class side method
myMeth
      ^self.
3. Add a new project in Iceberg and add this package.
4. Commit.
Normally here everything goes well.

5. modify the definition of trait TMyTest2
6. Trait << #TMyTest2
        traits: {TMyTest1};
	package: 'MyPackage-Exception'
7. Commit
Normally the error raises here
image

I don't know if it works in Pharo 11. 
The way you define the trait is important.
The fact that the trait you are modifying already got a class side method is important because adding the used traits has an impact on the trait definition and the class trait definition and it is important that you modify it and not that you add it.

guillep:
Fixes #1608 

Class side definitions should be created always, for traits as well as for classes.

### DIFF
--- a/Iceberg-Tests/IceClassesCherryPickingTest.class.st
+++ b/Iceberg-Tests/IceClassesCherryPickingTest.class.st
@@ -39,7 +39,7 @@ IceClassesCherryPickingTest >> testAddClassDependingInATrait [
 
 	self assert: diff children size equals: 1.
 	
-	self assert: (diff / self packageName1) children size equals: 3.
+	self assert: (diff / self packageName1) children size equals: 4.
 	self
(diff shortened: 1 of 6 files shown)


guillep:
Tests are green, except for some timeouts.

guillep:
@anneetien confirmed it works in her original Moose issue

PR #1610, opened 2022-09-02
Only compute critiques when needed

sonibla:
There's a checkbox to prevent running critiques on commit, but its value was only used to enable/disable the critiques **display**, not its computation (IceTipCritiquesBeforeCommitBrowser>>#openIfCritiques)

I added logic in IceTipCritiquesBeforeCommitBrowser>>#updatePresenter to also enable/disable the computation


Ducasse:
Tx for the catch!

PR #1611, opened 2022-09-06
Fix/1426: ZnInvalidUTF8 when reading a commit message

### LINKED ISSUE #1184: IceGitHubError: Not found on cloning private github repos
When I clone a private GitHub repo using Iceberg using HTTPS the Iceberg tool asks for credentials and then clones correctly into a local repo in filesystem - but throws a "NotFound" error in Pharo afterwards.
One then always have to use the "attach existing" afterwards to find the already cloned repo and connect to it. 

Reason is because for private repos one gets a HTTP 404 that does not pass the #isSuccess test in responseWithValidationDo:

### LINKED ISSUE #1426: ZnInvalidUTF8 when reading a commit message
The error message is `ZnInvalidUTF8: Illegal leading byte for utf-8 encoding`.

This happens in a P9 with Iceberg v2.0.3 . Reproduce with:

- Open Iceberg repositories list
- Clone https://github.com/pharo-graphics/Bloc/
- Open the bloc repository (cmd+r)

The error is in this commit: https://github.com/pharo-graphics/Bloc/commit/8237f3e5ed9f8100639419e5b677a2b97cc54c55

guillep:
Fix #1426
Fix #1184

- upgrade libgit for better encoding management of messages
- do not block repository creation if github API does not answer (not found, or missing credentials)

### DIFF
--- a/Iceberg-Plugin-GitHub/IceTipGitHubRepositoryPanel.class.st
+++ b/Iceberg-Plugin-GitHub/IceTipGitHubRepositoryPanel.class.st
@@ -23,11 +23,17 @@ IceTipGitHubRepositoryPanel >> configureBuilder: aBuilder [
 { #category : #'querying - github' }
 IceTipGitHubRepositoryPanel >> getGitHubRepository [
 
-	^ [ IceGitHubAPI new
-		beAnonymous;
-		getRepository: self userName project: self projectName ]
-			on: LGitNoCredentialsProvided
-			do: [ nil ]
+	^ [ [ IceGitHubAPI new
+		  beAnonymous;
+		  getRepository: self userName project: self projectName ]
+		  on: LGitNoCredentialsProvided
+		  do: [ nil ] ]
+		  on: IceGitHubError
+		  do: [ :error | 
+			"If not found, probably means a private repository => ignore it"
+			error messageText = 'Not Found'
+				ifTrue: [ ^ nil ].
+			error pass ]
 ]
 
 { #category : #initialization }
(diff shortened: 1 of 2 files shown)


Iceberg-Plugin-GitHub/IceTipGitHubRepositoryPanel.class.st
```diff
@@ -23,11 +23,17 @@ IceTipGitHubRepositoryPanel >> configureBuilder: aBuilder [
...
-			do: [ nil ]
+	^ [ [ IceGitHubAPI new
+		  beAnonymous;
+		  getRepository: self userName project: self projectName ]
+		  on: LGitNoCredentialsProvided
+		  do: [ nil ] ]
+		  on: IceGitHubError
+		  do: [ :error | 
+			"If not found, probably means a private repository => ignore it"
+			error messageText = 'Not Found'
```
astares, line 34: Can we have a better message text than "Not found". The user might wonder what is not found and how he can proceed

guillep: This is the message we have from the github request, so I don't think I have access there...
Anyways, I know the solution is sub optimal, so if you have any ideas I'm a taker ! :)

astares: My fault - so small on the phone that I though it is an assignment, not comparision. Yes you can then hardly change it - thanks for the feedback!

astares (approved):

PR #1627, opened 2022-11-17
Fixes #1571, commit button not always visible

### LINKED ISSUE #1571: Iceberg commit browser not showing all contents when opened
**Bug description**
Iceberg Commit Browser hides some of its contents, to show the contents, users must manually resize

**To Reproduce**
1. Go to Iceberg
2. Click on button 'Commit' to open Iceberg Commit Browser
3. Resize the Commit Browser and you will see the hidden contents

**Expected behavior**
The Commit browser should display all of its contents when it is opened

**Screenshots**
Commit browser as shown when is opened:

image

Commit browser after manually resizing
image

kasperosterbye:
Fixes #1571 (sort of)
Move commit button to left side

### DIFF
--- a/Iceberg-TipUI/IceTipCommentPanel.class.st
+++ b/Iceberg-TipUI/IceTipCommentPanel.class.st
@@ -122,7 +122,7 @@ IceTipCommentPanel >> initializePresenters [
 		yourself.
 
 	actionButtonBar := self newButtonBar.
-	actionButtonBar add: (commitButton := self newButton
+	actionButtonBar placeAtStart; add: (commitButton := self newButton
 		label: 'Commit';
 		icon: (self iconNamed: #smallOk);
 		help: ('Commit your changes to {1}' format: { self repositoryName });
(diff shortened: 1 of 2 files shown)


kasperosterbye:
I have no idea what is the source of those errors, but if they are related to the change I made it is actually somewhat impressive.

kasperosterbye:
Despite all the failing errors, but I added a the same "fix" to `IceTipPreviewBrowser` (which then applies it to its three subclasses).

estebanlm:
hi @kasperosterbye. 
 I am sorry, but this is not a real fix, is a patch :) 
I cannot accept it because it will break all the UI design of iceberg (where ALL action buttons are in bottom right corner). 
While this is a decision that could be discussed (why not at the middle?, why not buttons occupying all space?...), this would imply to change the whole tool.
Now, why this problem was happening? Well, because the addition of "Run critics while committing" choice put stress on the width of the "push" toolbar actually pushing out the whole panel, including the commit button.
This is partially a bug on Spec, but also a problem on the choice of how to present the information there. 
I made a change and it will now look like this: 

Screenshot from 2022-11-24 10-18-45

I replaced the checkboxes by toggle buttons, saving in practice the space needed to not push the button out.
This will fix the commit push button problem while not changing the UI design ;)

Now I need to decide if I will change also the "push" checkbox...

kasperosterbye:
I like your design!
I know it is a patch, and just thought it might be useful until a real solution was available.

PR #1632, opened 2022-11-24
Adapt to latest spec2 pass2

estebanlm:


estebanlm:
this PR fixes https://github.com/pharo-vcs/iceberg/issues/1571

Issue #1637, opened 2022-12-01
Adding a SSH credential from UI raises an Error


hernanmd:
In Pharo 11, adding a SSH credential from UI raises an Error after clicking the browse Public or Private buttons to select the Public/Private jeys.

How to reproduce

- `IceTipCredentialsStoreBrowser new open.`
- Click in "+ SSH Pair" button.
- Click in the folder button right to the Public Key or Private Key.

If the parameter passed as location in Iceberg side (received as path in UITheme) is nil then #chooseReference will raise an Error with a Debugger because path is asked #asFileReference.

PR #1638, opened 2022-12-01
Fix for adding a SSH credential from UI raises an Error

hernanmd:
Instead of nil, this modification defaults to the HOME FileReference in IceTipLocationPresenter>>location, avoiding an Error when a SSH credential is added from UI. This happens after clicking the browse Public or Private buttons to select the Public/Private keys.


hernanmd:
This fixes #1637 feel free to review.

PR #1642, opened 2022-12-06
Fix Crash when invalid credentials

guillep:
Related to: https://github.com/pharo-project/pharo-vm/issues/513

When the credentials callback opens the dialog, if an error happens during the dialog opening, libgit crashes.
This PR:
 - updates libgit to v3.0.3 that does a better management of credentials callbacks
 - updates iceberg to not fail when opening the pop up


estebanlm:
CI is failing because of some problem on the vm version (I was trying to fix that). I will accept the change and hope for the best in the real CI ;)

PR #1649, opened 2022-12-16
Use ED25519 keys as default instead of RSA to improve security and follow new GitHub standards

### LINKED ISSUE #1648: Use ED25519 keys instead of RSA to improve security and follow new GitHub standards
Details see https://github.com/pharo-project/pharo/issues/12115

astares:
Fix #1648


guillep:
Thanks for taking care about this, it is important.

estebanlm:
problem seems more complex than this... we need to check better.

guillep:
> problem seems more complex than this... we need to check better.

Sure, but making these the default is a good start!
I imagine that this should be tested on different machines (libssh/libssl on nixes, in windows libgit should be compiled to use the system's encryption mechanisms).

@astares have you tested this with ed25519 keys? If yes, on what platform/version and Pharo version?

estebanlm:
no, because we will make it the default and it will not work ;)
yesterday in discord channel was stated that this is not working.

guillep:
But that's why I say it should be tested. Please, read all my message :/

estebanlm:
and it was tested, and not working ;)
also I said that, so we are both saying the same :)

astares:
@estebanlm RSA is also not working - at least not on Windows 
 
 Additionally it is seen as insecure and GitHub uses now ed25519 as default since a longer time.

estebanlm:
RSA is working on windows, I just checked that.
(but new added RSA keys may not work, as it was pointed at discord)

astares:
@estebanlm 
If we speak about people who have already an old RSA setup, For them it might be working. 

All the poor souls

- who do not have an "aged" RSA key
- reset maybe their environment due to hardware change without keeping the old key (and just regenerate)
- and especially all the poor newbees
who will be hit by this will be screwed / dissapointed.

estebanlm:
but there is a difference between "RSA does now works because github does not wants" and "RSA does not works because iceberg does not wants"
and then, the fact RSA is working means there is something specific to dig into it.

astares:
@estebanlm 
If we speak about people who have already an old/ancient RSA setup, For them it might be working. 

But all the Pharo users
- who do not have an "aged" RSA key
- who reset their environment (due to hardware change or OS resetup) without keeping the old key (and just regenerate)
- and especially all the poor newbees
will immediately hit by this

And I know it requires care and testing beside so many other also important tasks.

jecisc:
@estebanlm Can you describe what is not working?

I’m now using ED25519 on my new machine updating this same setting and it works. But I’m on OSX, so maybe the problem is on windows?

jecisc:
Esteban confirmed that it was a windows problem

astares:
Might sound strange but could it be some path separator issue again?

gjvanderheiden:
Pharo 10 doesn't work (after setting the new keyfiles in Iceberg settings)
Pharo 11 is works, but crashes at many other points (not workable)

Mac os X, Intel. 
step1 settings
step 2 clone
result- error code -20

estebanlm:
after debugging a lot and search a lot, I found this: 
https://github.com/libgit2/libgit2/security/advisories/GHSA-8643-3wh5-rmjq

windows is using libgit2 1.4.4 (my linux, 1.5.1). 

may this be related @tesonep ?

gjvanderheiden:
Pharo 11 (Mac os X)  also uses 1.4.4 and works fine. Pharo 11  also got the default public and private key values to ED25519, due to tis PR I guess.

Simple compiling git2 lib and stuck it in the plugin folder doesn't fly. (Also changed the file name in the code where the debugger pops up)

Upgrading to 1.4.5 doesn't hurt either I'd say. libgit2 recommands that.

Edit: Correction: Downloaded a new version Pharo 11, doesn't work. Complains about invalid keys

astares:
@tesonep is adding 1.5.1 for Windows. 

See https://github.com/pharo-project/pharo-vm/commit/4784ae168e410baf2ec90cceff44a7d79994933e

estebanlm:
yeah, update to 1.5.1 is good, but it does not fixes the problem (still looking for)

gjvanderheiden:
Pharo 11 does work, I goofed with the settings. (gitlab vs github cough cough). Launching with the Pharo launcher doesn't give me de default ed25519 keys yet, they where stored in a setting file. 

I also tried using the libgit2 from GlamorousToolkit in Pharo 10. Gt is also Pharo 10 based. No difference. Gt also works. That's my work around right now. Start Gt directly click on Morphic World, which gives me a Pharo 10 image. Can't work with Coder in Gt, not user friendly in my opinion.

PR #1672, opened 2023-01-30
Fix for issue #11969 in pharo repository: Better description "packages unloaded packages

hernanmd:
Add a context group menu for the dialog in IceTipAddNewPackagesDialog. Implements two options: Select All and Unselect All packages.
These are implemented in two IceTipSelectPackagesCommand subclasses.

Reported in: https://github.com/pharo-project/pharo/issues/11969

### DIFF
--- a/Iceberg-TipUI/IceTipSelectAllPackagesCommand.class.st
+++ b/Iceberg-TipUI/IceTipSelectAllPackagesCommand.class.st
@@ -1,0 +1,50 @@
+"
+Implements the select all option behavior in the ""Select New Packages to Load"" Iceberg dialog.
+
+"
+Class {
+	#name : #IceTipSelectAllPackagesCommand,
+	#superclass : #IceTipSelectPackagesCommand,
+	#category : #'Iceberg-TipUI-Commands'
+}
+
+{ #category : #default }
+IceTipSelectAllPackagesCommand class >> defaultDescription [
+
+	^ 'Opens a dialog to choose select and/or removing new package(s) from the repository'
+]
+
+{ #category : #default }
+IceTipSelectAllPackagesCommand class >> defaultName [
+
+	^ 'Select all packages'
+]
+
+{ #category : #testing }
+IceTipSelectAllPackagesCommand >> canBeExecuted [
+	" Answer <true> if the receiver has items to display and they are not all selected "
+
+	| pkgList |
+
+	pkgList := self context packagesList.
+ 	^ pkgList items notEmpty and: [ pkgList items size ~= pkgList selectedItems size ]
+]
+
+{ #category : #executing }
+IceTipSelectAllPackagesCommand >> execute [
+	" Private - Select all the receiver's packages "
+
+	self packagesList items
+		do: [ : pkgItem |
+			self packagesListSelectionColumn
+				select: [ : c | c isKindOf: SpCheckBoxTableColumn ]
+				thenDo: [ : c | c onActivation cull: pkgItem ] ]
+		displayingProgress: [ : pkgItem | 'Selecting ' , pkgItem name ].
+	self packagesList refresh
+]
+
+{ #category : #accessing }
+IceTipSelectAllPackagesCommand >> iconName [
+
+	^ #checkboxSelected
+]
--- a/Iceberg-TipUI/IceTipSelectAllPackagesCommand.class.st
+++ b/Iceberg-TipUI/IceTipSelectAllPackagesCommand.class.st
@@ -26,7 +26,7 @@ IceTipSelectAllPackagesCommand >> canBeExecuted [
 
 	| pkgList |
 
-	pkgList := self context packagesList.
+	pkgList := self packagesList.
  	^ pkgList items notEmpty and: [ pkgList items size ~= pkgList selectedItems size ]
 ]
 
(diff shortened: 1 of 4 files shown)


astares:
Thanks! You PR already adds a select all button - very nice.

But additionally I wrote in the issue that the **message displayed is also confusing**
because I do not understand why this dialog really came up.

We should ideally have a **more descriptive and easy text** that tells 
 - what happened
 - what the user now needs to do
 
Or maybe I'm the only one and just too stupid to understand what is written on the dialog
and why it popped up and which packages to choose to solve something

hernanmd:
> Thanks! You PR already adds a select all button - very nice.
> 
> But additionally I wrote in the issue that the **message displayed is also confusing** because I do not understand why this dialog really came up.
> 
> We should ideally have a **more descriptive and easy text** that tells
> 
>     * what happened
> 
>     * what the user now needs to do
> 
> 
> Or maybe I'm the only one and just too stupid to understand what is written on the dialog and why it popped up and which packages to choose to solve something

Feel free to propose here a more descriptive text and I will commit it. I agree that the current message is confusing.

Iceberg-TipUI/IceTipSelectAllPackagesCommand.class.st
```diff
@@ -0,0 +1,50 @@
...
+{ #category : #executing }
+IceTipSelectAllPackagesCommand >> execute [
+	" Private - Select all the receiver's packages "
+
+	self packagesList items
+		do: [ : pkgItem |
+			self packagesListSelectionColumn
+				select: [ : c | c isKindOf: SpCheckBoxTableColumn ]
+				thenDo: [ : c | c onActivation cull: pkgItem ] ]
+		displayingProgress: [ : pkgItem | 'Selecting ' , pkgItem name ].
```
guillep: This method looks fishy:
 - is `c isKindOf: SpCheckBoxTableColumn` the only way to select all checkboxes in a colum in a table? Maybe @estebanlm knows better, I cannot help here
 - Why a progress bar here? How many packages do we need in a repository so that this is SOOO slow that the progress bar becomes actually visible and gives reasonable feedback? :)

Iceberg-TipUI/IceTipSelectPackagesCommand.class.st
```diff
@@ -0,0 +1,24 @@
...
+	^ self context packagesList
+]
+
+{ #category : #executing }
+IceTipSelectPackagesCommand >> packagesListSelectionColumn [
+	" Answer a <SpCheckBoxTableColumn> to select packages in the receiver's context package list "
+
+	| compositeCol |
+	compositeCol := self packagesList columns.
+	^ compositeCol anyOne columns select: [ : c | c isKindOf: SpCheckBoxTableColumn ]
```
guillep: Same as above for the isKindOf:, is this the way in Spec @estebanlm ?

Iceberg-TipUI/IceTipUnselectAllPackagesCommand.class.st
```diff
@@ -0,0 +1,50 @@
...
+{ #category : #executing }
+IceTipUnselectAllPackagesCommand >> execute [
+	" Private - Unselect all the receiver's packages "
+
+	self packagesList items
+		do: [ : pkgItem |
+			self packagesListSelectionColumn
+				select: [ : c | c isKindOf: SpCheckBoxTableColumn ]
+				thenDo: [ : c | c onDeactivation cull: pkgItem ] ]
+		displayingProgress: [ : pkgItem | 'Unselecting ' , pkgItem name ].
```
guillep: And here

guillep (changes requested):
I like the idea. I don't know if this is the best (or the only) way to do this.
- I would prefer @estebanlm 's opinion on checkbox thing
- @astares could you propose a new text description? I see in the original issue you proposed a grammatically better sentence, but it's not clear if you wanted to go further

Iceberg-TipUI/IceTipSelectAllPackagesCommand.class.st
hernanmd: > * Why a progress bar here? How many packages do we need in a repository so that this is SOOO slow that the progress bar becomes actually visible and gives reasonable feedback? :)

IMO selection is slow because of the current implementation, which is the best I could arrive with the current Spec 2 API.  So we should measure here, but if there's a better way to select items I'm looking forward to change it.

guillep: Let's wait for @estebanlm's input.
Maybe it's slow because it's redrawing the entire table for each selection?
Maybe we need a new API to select many things at the same time?

estebanlm:
yes, this is not cool. 
essentially, you do not check all checkboxes by modifying the UI (which will be like simulating you pressing the buttons). Instead, you need to check the model on which the table is based and later call an updatePresenter (which will change the UI just one time).
In this particular case, you should first add all packages to `selectedPackages` and the call updatePresenter, which will do the trick.

estebanlm (changes requested):
please, fix the way the checkboxes are updated

hernanmd:
> please, fix the way the checkboxes are updated


Thanks for the feedback. I've updated the implementation with your suggestion and it was a lot easier.
It can be tested with:

```smalltalk
(IceTipAddNewPackagesDialog on: RPackageOrganizer default packages) open
```

guillep (approved):
Yes, much better.
Thanks @hernanmd for the good energy.
Thanks @estebanlm for the super hints :)

astares:
@guillep / @hernanmd / @estebanlm 

Like Guille I appreciate all your help and efforts here too. 

But the **description** (which is still the old) is absolutely not helpful - and I can not propose a new one as long as I do not know what the dialog solves/ what the dialog is about (or what situation leads to making it necessary for Iceberg to pop it up).

You can call me stupid but I still fail to understand 
 - what kind of situation one has to get in where Iceberg needs to present this dialog. Can you give an example?
 - why do "packages unload packages" as it is mention in the description? 

The dialog text 

   _"One or more packages unloaded packages (...) has been changed."_

is not even a regular english sentence. 

Also the context menu on the description text is strange - the user is given the impression he can paste in here somthing
(which I think is a Spec2 issue):

image

Sorry for the rant - but from a user perspective there is still room for improvement.

PR #1673, opened 2023-02-02
Prevent recursive retries on createIcebergRepositoryWithFallbackFor:url:

hernanmd:
Prevent recursive retry on createIcebergRepositoryWithFallbackFor:url:
Fix for adding a SSH credential from UI raises an Error (https://github.com/pharo-vcs/iceberg/pull/1638)


estebanlm:
we need to discuss this. 
the retry was there for something :)

guillep:
> we need to discuss this. the retry was there for something :)

Yes, but this already retries: once.
The thing is that with the "retry" send there, this loops in an infinite recursion.

guillep:
I'll merge as these infinite recursions create a very bad impression.
We can enhance it later.

PR #1680, opened 2023-02-21
1279 merge menu should indicate the sense of the action

Ducasse:
fix menu and update of the dialog box so that we know in which branch we are merging into

### DIFF
--- a/Iceberg-TipUI/IceTipMergeCommand.class.st
+++ b/Iceberg-TipUI/IceTipMergeCommand.class.st
@@ -16,7 +16,7 @@ IceTipMergeCommand class >> defaultDescription [
 { #category : #'accessing - defaults' }
 IceTipMergeCommand class >> defaultName [
 
-	^ 'Merge into current'
+	^ 'Merge'
 ]
 
 { #category : #testing }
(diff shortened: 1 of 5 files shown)


estebanlm:
this PR actually breaks a couple of tests.
and no idea if that test is correct or not (I am not friend of testing position of buttons and that, since they can change without breaking an UI) :(

Iceberg-TipUI/IceTipMergeCommand.class.st
```diff
@@ -10,13 +10,13 @@ Class {
...
 
-	^ 'Merge your current branch with another.'
+	^ 'Merge a given branch into your current branch.'
 ]
 
 { #category : #'accessing - defaults' }
 IceTipMergeCommand class >> defaultName [
 
-	^ 'Merge'
+	^ 'Merge into current'
```
estebanlm: and is this thing what is making tests fail. 
also, merges are always "into current", I do not see the reason to put this explanation :)

Ducasse: You should go and teach more often to students. It was a cheap way that people do not get confused. There are plenty of things that are obvious to you that are not for many people and also plenty of things that are obvious for other people that you do not have a clue. This is why a nice user interface helps people to learn. 
Now if tests break because we change a menu title then we have bad tests.

Issue #1681, opened 2023-02-21
Committing with empty description can lead to problems


Ducasse:
In my imagine I did 

```
IceTipEntityModel >> shortDescription

	^ self description asString lines ifNotEmpty: [ :f | f first ] ifEmpty: [  'No message' ]
```

to be able to work

PR #1682, opened 2023-02-21
Fixes: Committing with empty description can lead to problems #1681

Ducasse:


demarey:
error not related to the fix

demarey (approved):

PR #1687, opened 2023-03-13
1686-SSH-Key-pair-should-not-ask-user-it-should-use-the-user-from-the-URL

### LINKED ISSUE #1686: SSH Key pair should not ask user, it should use the user from the URL
If the URL is git@github.com:user/project the user to use is git. 
If there is another user, the URL should be in sync.

tesonep:
The username of the SSH credential should come from the SSH URL not from the configuration.

- Fixing the presenters
- Fixing cancel when editing.

Fix #1686


estebanlm:
I have valid failing tests @tesonep :)

Ducasse:
So what do we do with this fix?

Ducasse:
This PR is too old. Should we close it?

Ducasse:
@tesonep what do we do with this PR?

PR #1688, opened 2023-03-15
fix: ensure right method side (class vs instance) is chosen when looking through history

### LINKED ISSUE #12999: history displays meta class method instead of instance method
Viewing history of an instance method can show the history of the class method if one exists with the same selector.

**To Reproduce**
Find a class that has both instance and class methods with the same selector. Here I picked `SmaCCScanner>>initialize` which has a class side `#initialize` method: 
image

**Version information:**
 - Pharo Version: Both 10 & 11

The `#definitionFor:path:commit:` methods need to compare the `classIsMeta` in addition to the selector: 
```
each selector = aMethod selector
						and: [ each classIsMeta = aMethod methodClass isClassSide ]
```
I tested on the IceLibgitTonelLog method, but IceLibgitFiletreeLog has most of the same logic and likely needs the same fix.

This was originally reported in https://github.com/feenkcom/gtoolkit/issues/3136.

hellerve:
Fixes https://github.com/pharo-project/pharo/issues/12999


Ducasse:
@estebanlm it looks like a good fix for P11.
But you decide.

PR #1702, opened 2023-06-06
remove methods that are the same in the superclass

MarcusDenker:
... trivial cleanup


pavel-krivanek (approved):
I checked all the methods, everything fine!

PR #1703, opened 2023-06-06
Cleanup-DeadCode-MCDefinition

MarcusDenker:
method not called, it itself sends a non-existing selector


Ducasse:
Change looks good.

PR #1713, opened 2023-07-16
First handling simple inform: and also those that were self applicati…

Ducasse:
…on inform: and simply invoke Object>>inform:


Ducasse:
I will stop for now because I do not know how to convert the remaining ones such as 
informUser:during: or choice with three forms such as trueChoice:falseChoice:cancelChoice:.
So do not hesitate to review and integrate, or drop it.

jecisc:
This change is quite invasive. 

This is not a problem of this PR I think but of the inform system. Before it was a little popup in the corner vanishing itself. Now we have a dialog in the middle of the screen that we need to close by interacting with it.

PR #1728, opened 2023-09-04
Unify snapshot method with Monticello

jecisc:
Iceberg implemented its own MCPackage>>snapshot method to modify the one of Monticello. 

The method of Monticello was updated to use the same change as Iceberg but also it was updated to reduce the number of categories (package-tag) manipulations. I want to check here if we could use this new version for Iceberg too as a step toward getting rid of the categories in Pharo


jecisc:
Done in:https://github.com/pharo-vcs/iceberg/pull/1732

PR #1732, opened 2023-09-06
Remove usage of categories in MCOrganizationDefinition

jecisc:
Base the organization definition on packages and tags instead of categories.

#ensureMCClassOrganisationHasCorrectStructure: has been simplified because now the organization knows the package and we do not need to do magic to ensure we have the right one

IcePackageDefinition has been updated to use the package name and tag names instead of categories.

This change includes https://github.com/pharo-vcs/iceberg/pull/1728


jecisc:
This is currently broken because of Tonel

jecisc:
The failing tests seems to work in the latest Pharo 12. Let's restart the CI

jecisc:
This PR is now working :)

PR #1734, opened 2023-09-08
Renamed all the xModel into xPresenter same for xDialog into xDialogP…

Ducasse:
…resenter just to help understanding the code.


Ducasse:
It is unclear to me if we are running the iceberg tests because there are not loaded by default.

guillep:
They are! click on the `Details` link on the github actions :)

imagen

Ducasse:
Thanks. This is an excellent news!

PR #1744, opened 2023-10-11
do not unload and reload tonel and spec while loading iceberg

demarey:


jecisc:
Done in another PR to fix the CI and remove Ring2 usage

PR #1747, opened 2023-10-19
Update Ring version

jecisc:
To follow Pharo 12 changes


jecisc:
I'm doing a new PR to remove the use of the external Ring

PR #1748, opened 2023-10-23
Adding support for Proxy URL

tesonep:
- Updating to Libgit 3.0.8 to have proxy for the fetch.
- Adding proxy configuration from the NetworkSettings
- Setting proxy in Clone, Fetch, and Pull
- Adding Tests


tesonep:
Fix https://github.com/pharo-project/pharo/issues/13726

PR #1765, opened 2023-11-21
Fix CI and remove Ring2 dependency

jecisc:
The goal of this PR is to fix the CI. It seems to work on my local copy of Iceberg, now let's see what the CI says. 

I removed the replacement of Spec and Tonel in the pre load script because this is not needed and buggy. I also removed the used of the external Ring for tests and use the Ring version in the Pharo image instead. 

I also removed some scripts and made sure all scripts were in the same place


jecisc:
In order to do this change I had to change things in Pharo itself to improve Ring and it seems that the changes did not made it into SmalltalkCI used images yet

MarcusDenker (approved):

PR #1767, opened 2023-11-21
Avoid useless garbage collects

jecisc:
Some garbage collects seems useless in Iceberg tests (I ran the tests multiple times without troubles) and are slowing down Iceberg tests


jecisc:
Tests passed and builds are faster with this change

MarcusDenker (approved):

PR #1776, opened 2023-12-05
1775 ice tip tag model does not understand branch name when browsing tags

astares:
Fix #1776


astares:
Two platforms green, one test fails on Windows #testFetchUpdatesTrackedRemoteBranch 
(but it is also failing for other PRs like https://github.com/pharo-vcs/iceberg/pull/1778) so it looks like a CI issue

image


and when I run this test manually with latest Pharo P12 on Windows with this PR here it is green:

image

guillep (approved):
Thanks!

PR #1782, opened 2023-12-16
Remove only the reverted diff entry rather than recompute the whole diff after revert (fixes #1745)

### LINKED ISSUE #1745: Avoid refreshing the whole tree after reverting a change in a commit browser
**Describe the request**
When I revert a change in an Iceberg commit browser, this triggers a full refresh of the change set—which for the core Pharo repository, on Windows, can easily take 10-15 seconds. Sometimes I'm experimenting and determine that I've gone down the wrong path, or have added a bunch of debugging/logging code I need to remove once I've figured out a problem. I'd like to revert them all quickly from a commit browser, but between the lack of multiselect (see #1206) and a full tree refresh each time, this is extremely slow.

**Expected behavior**
It should be possible to remove only the reverted entry from the tree, without re-diffing, speeding up the process dramatically. This does mean that if any other changes have happened since the commit browser was last refreshed, they will not be caught up—but this is already the case if you were to make other changes and _not_ revert, so really it needs to be up to the user to refresh before actually committing. (Or if we could track "this commit browser might be outdated" and warn, that would be the best—separate issue.)

**Expected development cost**
I did something that works but is a bit ugly, maybe it can be cleaned up enough to be workable? WIP PR inbound.

**Version information:** Any OS, any Pharo

daniels220:
This feels pretty brittle and like the model objects should be able to do more of the work—for instance, the fact that I have to separately check `isNoModification or: [isExtensionDefinition]`, because extension class definitions are always "added", yet have no content of their own, where a packaged class definition could be added as just an empty class. Also little things like the tree entries not having a "remove" API so I have to traverse up and ask the parent to remove the child, etc. Feels like the API here is under-developed, basically—and that makes sense because we ultimately could do things like update the diff in-place when a method is saved (after all, the only thing that can possibly change is the diff entry for that method...), which would probably drive the creation of a richer API here.

So—if this seems like a step forward, great. If you have any suggestions for making it less ugly and brittle without tackling a major revamp of the diff panel, I'm happy to put in a little more work.

### DIFF
--- a/Iceberg-TipUI/IceTipDiffPanel.class.st
+++ b/Iceberg-TipUI/IceTipDiffPanel.class.st
@@ -84,6 +84,29 @@ IceTipDiffPanel >> diffContentsLeft: leftString right: rightString [
 		rightText: rightString
 ]
 
+{ #category : 'initialization' }
+IceTipDiffPanel >> diffEntryRemoved: announcement [
+
+	| parent |
+	(parent := announcement entry parent) removeChild: announcement entry.
+	"Removing all changes to a class should remove the class node too,
+	unless the class definition has changed as well. Extension definitions
+	may appear as an 'addition' or 'removal', but can only exist when methods
+	of that class have changed. Repeat this process to check if the containing
+	package no longer has any changes. Cannot remove the root node as it has
+	no parent, but it's fine to just leave it there empty--though perhaps we should
+	close the browser instead in that case?"
+	[
+	parent parent notNil and: [
+		parent isEmpty and: [
+			parent value isNoModification or: [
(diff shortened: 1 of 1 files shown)


Ducasse:
Did you run the tests?

daniels220:
Which tests are you referring to? It doesn't look like there were any test failures in the CI build (the one Windows issue seems unrelated and I've seen before in other CI runs).

If what you mean is that I should write tests, fair. I I only just went looking and found that there _are_ existing tests of the UI, as they're not loaded by default in a Pharo distribution image (reasonable enough), I had to load them from the repo. So I'll get to that when I have a chance...

guillep (approved):
Looks good enough, let's try it! Thanks @daniels220 !

PR #1784, opened 2023-12-16
Hide "Reset <branch> to <commitish>" when a tag is selected (fixes #1783)

### LINKED ISSUE #1783: Expanding tag list gets IceTipTagModel DNU #branchName
Opening Repository view (`IceTipRepositoryBrowser`) on a repo with tags and clicking the disclosure triangle to view them gives `IceTipTagModel` DNU `#branchName`, from `IceTipResetBranchToThisCommitCommand>>defaultNameFromContext`. If you close the debugger and try to right-click in the commit list, it happens again. If you explicitly select a branch, the error goes away.

Looking at `#defaultNameFromContext`, the sibling `IceTipBranchModel` does understand this message, and clearly the code was written assuming `selectedBranch` would return one of those. It seems like `self selectedBranch` is whatever is selected in the sidebar, and `selectedCommitish` is what's selected in the main panel. But the sidebar selection might not _be_ a branch, and then we have a problem.

I think the problem here is that it isn't really valid/reasonable to say "reset <tag-name> to <commit>". You can do something like it, but you wouldn't use `git reset` on the command-line, you'd use `git tag -f`. And indeed `IceTag` has no ability to modify what commit it points to, and creating one is done externally by `IceGitCommit>>createTag:`. So the command should just be disabled or absent from the menu when a tag is selected instead of a branch. If we wanted to have a separate "move tag <tag-name> to <commit>", we could, but it would need to be written basically from scratch using a call to `createTag:targetId:force: true`.

After some poking around I think I figured out the right way to do this, PR incoming.

daniels220:


guillep (approved):
Thanks!

PR #1789, opened 2024-01-18
First step towards multiple working copies

guillep:
The current repository-centric design allows to have only single working copy.
Indeed, before the repository used to do

```smalltalk
Repository >> foo
   self workingCopy foo
```

This PR moves several responsibilities from the repository to the working copy object.
A lot of code gets clearer, allows to have multiple working copies, and as backwards compatibility the repository still has a default working copy as before.

Other changes:
 - IcePackage knows the working copy and not the repository (this is better because a package may exist in a commit and not on another)
 - Some UI models got updated to work on the working directory
 - I renamed/aliased some very generic methods that made coding very difficult (model, entity => repositoryModel, workingCopyModel, repository, workingCopy). The generic names imply a false polymorphism. There is no polymorphic usage between a workingCopy and a repository except for asking for a `description`. The uses I stumbled upon use now the less generic version that correspond on each situation.
 - removed `IceTipCache` et al that complicated debugging and modifications for not much


Ducasse:
@guillep there are lot of tests of iceberg that broke.

Ducasse:
Do you want that I give a try to get some corner cases just using the UI?

guillep:
If you want, but just FYI, I redid all these changes from scratch so this is much cleaner than what we drafted yesterday.
I'm much more confident on this PR than on what we did yesterday ^^.

Ducasse:
Ok I do not have time right now so once this is integrated I will use it and we will see :)

PR #1753, opened 2023-10-27
Make repair detached display the current branch

bouraqadi:
Allow repair detached by "merging" into the current branch on disk. We simply display the current branch and mark it and "current". This is useful for example after changing some resource file on disk.
This fixes issue listed in Pharo https://github.com/pharo-project/pharo/issues/6797

PR #1791, opened 2024-01-18
Revert 1790 revert 1753 6797 issue from pharo

guillep:
Redo https://github.com/pharo-vcs/iceberg/pull/1753 on dev-20 branch


guillep (commented):
Oh no the merge did not work as expected :( I think it's better to just cherry pick the changes

PR #1792, opened 2024-01-18
Allow repair detached by "merging" into the current branch on disk

guillep:
Redo https://github.com/pharo-vcs/iceberg/pull/1753 on dev-20 branch

We simply display the current branch and mark it and "current"


guillep:
Now merge is good, thanks @bouraqadi and @demarey !

PR #1793, opened 2024-01-18
Implement IceWorkingCopyPatcherVisitor>>visitExtensionDefinition:

### LINKED ISSUE #14395: Error reverting all changes to class in Iceberg commit window
**Bug description**
If I have several changes to extension methods of a class, and try to revert all 
'IceWorkingCopyPatcherVisitor does not understand #visitExtensionDefinition:'

**To Reproduce**
1. Make several changes to extension methods of the same class, in the same package.
2. Open a commit browser for the repo containing that package. The method changes are grouped under a single tree node for the class.
3. Right-click the class and choose Revert
4. See `IceWorkingCopyPatcherVisitor does not understand #visitExtensionDefinition:` error

**Expected behavior**
Revert all changes to extension methods in that class for that package (the nodes under it in the tree).

**Version information:**
 - OS: Windows/any
 - Pharo Version 11

**Expected development cost**
I imagine the solution is implementing the method in question and shouldn't be too difficult. I'm not familiar with Iceberg, so I'm probably not the best person to do this, but if it annoys me enough I might try.

daniels220:
Fixes error trying to revert all extension-method changes in a class, i.e. right-click on the class extension "definition" node and choose revert. Extension definitions are just a container and have no content of their own, so the same implementation as #visitDirectoryDefinition:--just visit the children--seems reasonable.

This fixes [pharo-project/pharo#14395](https://github.com/pharo-project/pharo/issues/14395)—when I reported it I didn't realize Iceberg had its own repo...

### DIFF
--- a/Iceberg/IceWorkingCopyPatcherVisitor.class.st
+++ b/Iceberg/IceWorkingCopyPatcherVisitor.class.st
@@ -90,6 +90,13 @@ IceWorkingCopyPatcherVisitor >> visitDirectoryDefinition: anIceDirectoryDefiniti
 ]
 
 { #category : 'visiting' }
+IceWorkingCopyPatcherVisitor >> visitExtensionDefinition: anIceExtensionDefinition [
+	"An extension definition represents no changes on its own, visit the individual extension-method changes."
+
+	self visitChildrenOf: currentNode
+]
+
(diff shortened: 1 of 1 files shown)


guillep (approved):
Yes!

Issue #1802, opened 2024-02-09
Regression: Default format of a repo should not be filetree


Ducasse:
In P12 filetree appears first and this is not good because it should be tonel

guillep:
Fixed in #1804

PR #1804, opened 2024-02-15
Default format

guillep:
Fixes https://github.com/pharo-vcs/iceberg/issues/1802

Set the default format in the format list in the edit project dialog when open.
Otherwise this is only set when selection changes.

imagen

Should this PR go against `dev-2.0` or `Pharo12`.
It seems `Pharo12` is ahead by several commits.


estebanlm:
I am working on Pharo12 to comply with what we agreed. For now the integration branch is still dev-2.0 so is ok

estebanlm:
but... I see that you fixed some problems I fixed too? (in Pharo12)
maybe better to go there, I am just waiting for the problems on SmalltalkCI to get fixed so I can finish this.

guillep:
Nope, all the commits there are yours!!! That's what I mean, there are several commits you integrated into Pharo12 instead of dev-2 and that's why I'm confused :)

PR #1808, opened 2024-02-25
Refactor IceTipTreeNode to support use of a FormSet

Rinzwind:
This pull request refactors IceTipTreeNode to support use of a FormSet (introduced in [Pharo pull request #14998](https://github.com/pharo-project/pharo/pull/14998)) for the icon and applies that to `#repositoryModelsByGroup` on IceTipRepositoryModel.

### DIFF
--- a/Iceberg-TipUI/IceTipRepositoryModel.class.st
+++ b/Iceberg-TipUI/IceTipRepositoryModel.class.st
@@ -397,7 +397,7 @@ IceTipRepositoryModel >> repositoryModelsByGroup [
 	^ {
 		  ((IceTipTreeRepositoryGroup on: self)
 			   name: 'Branches';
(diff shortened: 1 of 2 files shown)


Rinzwind:
There are two additional commits that I did not include in this pull request: commits a9abd57399489bee and 9fd55052324e1cf4. Commit a9abd57399489bee would change the SpImageTableColumn’s answer for `#readObject:` from a Form to a FormSet. That would not be a problem for `#visitImageColumn:` on SpMorphicTableCellBuilder as it just sends the answer `#asMorph`. But other uses of the SpImageTableColumn could break. I’m guessing that’s at least the case for [‘Spec-Gtk’](https://github.com/pharo-spec/Spec-Gtk).

PR #1829, opened 2024-06-24
Cut dependencies of Iceberg on UI

jecisc:
- Remove #morphicOpen: that is dead code
- Move IceMergeResolationRequest>>defaultAction to Iceberg-TipUI since it is here to open the Tip UI
- Move IceExperimentalFeature to Iceberg since this is raised in Iceberg's code and not in the UI code


jecisc:
Also fixing a pre load script for P13

PR #1830, opened 2024-06-25
Remove usage of the Snapshot Browser

jecisc:
The snapshot browser currently have only one user in Pharo that is the IceTipBrowseRepositoryPackageVersionCommand.

I propose to remove this command because we already have ClyBrowseIcebergCommitCommand that allows to do the same thing but in a better browser. With this, we will be able to remove yet another Morphic browser.


jecisc:
That is not so easy because apparently it's also used in other places :(

PR #1845, opened 2024-09-13
Remove FastTable #isRoot methods

jecisc:
Those methods are already in Pharo and Iceberg clashes with that so we should just remove them (They are identic)


jecisc:
CI is failing because here is what is happening:
- Pharo is bootstraped with the #isRoot method 
- Iceberg is loaded and overrides the #isRoot method 
- In the Iceberg CI we unload Icebger that remove the #isRoot mehtod
- The CI reload the new Iceberg where I removed the #isRoot method and the one of Pharo is not there anymore because of step 2

So the PR fails but it will work in Pharo

hernanmd (commented):
Looks good

PR #1846, opened 2024-09-28
Add method history browser context menu item for browsing the selected commit in the repository browser

Rinzwind:
When looking at the history of a method, I often want to see the wider context: what else was changed in the same commit, and what commits were done before and after. To support that, this pull request extends the commit list context menu in the window showing the history of a method with a menu item ‘Browse commit’ that opens a repository browser on the corresponding repository and selects the same commit. The first of the following two screenshots shows an example of using the menu item and the second one shows the repository browser that is opened:

<p align="center">

</p>

<p align="center">

</p>

### DIFF
--- a/Iceberg-TipUI/IceTipBrowseCommitCommand.class.st
+++ b/Iceberg-TipUI/IceTipBrowseCommitCommand.class.st
@@ -1,0 +1,28 @@
+"
+Command that opens a Repository Browser on the selected commit.
+"
+Class {
+	#name : 'IceTipBrowseCommitCommand',
+	#superclass : 'IceTipCommand',
+	#category : 'Iceberg-TipUI-Commands',
+	#package : 'Iceberg-TipUI',
+	#tag : 'Commands'
+}
+
+{ #category : 'default' }
+IceTipBrowseCommitCommand class >> defaultName [
+
+	^ 'Browse commit'
+]
+
+{ #category : 'execution' }
+IceTipBrowseCommitCommand >> commit [
+
+	^ self item commit
+]
+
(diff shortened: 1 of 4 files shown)


Rinzwind:
The speed of browsing larger repositories, such as the [‘pharo’ repository](https://github.com/pharo-project/pharo) itself, needs some improvement for this feature to be (more) useful though.

hemalvarambhia:
Is there a way to get the checks green? Why are they red?

Rinzwind:
I didn’t investigate beyond determining that the failing checks are probably not related to the changes in this pull request: for the [run of ‘Iceberg-Pharo-Integration’](https://github.com/pharo-vcs/iceberg/actions/runs/11086865761), the test that failed also failed in the [corresponding run for commit bd041b9](https://github.com/pharo-vcs/iceberg/actions/runs/11071485205), and for the [run of ‘Complete Test’](https://github.com/pharo-vcs/iceberg/actions/runs/11086865763), the file deletion error in the Windows job likewise affected the Windows job in the [corresponding run for commit bd041b9](https://github.com/pharo-vcs/iceberg/actions/runs/11071485210) (though in the latter it occurred while testing whereas in the former it occurred while loading).

Ducasse:
There is this test failing 
#testNoEquivalentSuperclassMethods apparently it is unique. 
Can you have a look? Else I will merge and I will fix it.

Rinzwind:
That test is the one that also failed in the [corresponding run for commit bd041b9](https://github.com/pharo-vcs/iceberg/actions/runs/11071485205) so I don’t think that’s related to the changes in this pull request.

PR #1847, opened 2024-10-14
Cut setting Morphic dependency

hernanmd:
Cut `IceTipCredentialsSettings class>>#settingsOn:` dependency on Morphic (needed by the new SettingBrowser). The change is compatible with both the old Setting Browser and the new Setting Browser.


hernanmd:
Failing test is not related since it was failing also in a previous build: https://github.com/pharo-vcs/iceberg/actions/runs/11086865761/job/30804911211?pr=1846

jecisc (approved):

PR #1849, opened 2024-10-18
merge P13 spec actions 

estebanlm:
this is already integrated, now I need to align the branches :)


estebanlm:
now errors are not mine :P

Iceberg-Plugin-GitHub/IceGitHubPullRequestDetailModel.class.st
```diff
@@ -172,7 +172,7 @@ IceGitHubPullRequestDetailModel >> newChecksTable [
 			evaluated: [ :aStatus | aStatus at: 'description' ];
 			yourself);
 		beResizable;
-		contextMenu: [ (self rootCommandsGroup / 'checks commands') beRoot asMenuPresenter ]
+		actions: (self rootCommandsGroup / 'checks commands')
```
tinchodias, line 175: Looks much better! 
I wonder if this affects Spec backends or they don't have impact to support this new API.
Have to check

estebanlm: you will need to implement it for toplo backend, there is no magic :P 
(but in morphic, this was a fairly easy implementation, I do not see why not on toplo)

PR #1855, opened 2024-11-18
Branch: CommandRepoBrowser1841,  Fixes # 1841

AlexisCnockaert:
idk if its the good way to implement it but it works


jecisc (changes requested):
This brings a menu with an Extra submenu that is not needed and an Action submenu that is empty.

Those should be removed

jecisc (approved):

jecisc (approved):

PR #1862, opened 2024-11-25
Fixes # 1861 on Pharo : location not a git repo when Importing existing clone

### LINKED ISSUE #1861: import from existing clone should be more robust
And it should not break when we select a location that it not a valid git repo.
It should either ask again or propose to cancel the action

Capture 2024-11-24 at 14 49 50

AlexisCnockaert:
Fixes #1861 
Actually maybe the **accept** method is a bit full now, but I think it's the easiest way to have the wanted behavior.
@Ducasse now when selecting a location that is not a git repo we just have this : 
Screen Shot 2024-11-25 at 14 25 03

### DIFF
--- a/Iceberg-TipUI/IceTipOptionDialogPresenter.class.st
+++ b/Iceberg-TipUI/IceTipOptionDialogPresenter.class.st
@@ -23,15 +23,11 @@ Class {
 IceTipOptionDialogPresenter >> accept [
 
 	[
-	self selectedType class = IceTipLocalRepositoryPanel
+	(self selectedType class = IceTipLocalRepositoryPanel and: [
+		 (IceRepositoryCreator isGitRoot: self selectedType location) not ])
 		ifTrue: [
-			(IceRepositoryCreator isGitRoot: self selectedType location)
-				ifFalse: [
-					self inform: ('"{1}" git repository not found' format:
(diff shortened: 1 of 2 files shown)


Iceberg-TipUI/IceTipOptionDialogPresenter.class.st
```diff
@@ -22,12 +22,22 @@ Class {
...
+			(IceRepositoryCreator isGitRoot: self selectedType location)
+				ifFalse: [
+					self inform: ('"{1}" git repository not found' format:
+							 { self selectedType location fullName }) ]
+				ifTrue: [
+					self doAccept.
+					self closeWindow ] ]
+		ifFalse: [
+			self doAccept.
+			self closeWindow ] ]
```
jecisc:
```suggestion
	[
	(self selectedType class = IceTipLocalRepositoryPanel and: [ (IceRepositoryCreator isGitRoot: self selectedType location) not ])
		ifTrue: [ self inform: ('"{1}" git repository not found' format: { self selectedType location fullName }) ]
		ifFalse: [
			self doAccept.
			self closeWindow ] ]
```

The code can be simpler

jecisc (changes requested):

jecisc (approved):

PR #1865, opened 2024-11-25
Fixes #1863 changed title desc for assertion failure

### LINKED ISSUE #1863: When importing existing clone, assertion failures should be more explicit
In IceTipLocalRepositoryPanel class, the validate method has titles for the pop-ups that are the same
Screen Shot 2024-11-25 at 14 46 44

AlexisCnockaert:
Fixes #1863 now we can dissociate more the different cases when we have cases of assertion failure

### DIFF
--- a/Iceberg-TipUI/IceTipLocalRepositoryPanel.class.st
+++ b/Iceberg-TipUI/IceTipLocalRepositoryPanel.class.st
@@ -90,5 +90,8 @@ IceTipLocalRepositoryPanel >> validate [
 	self
 		assert: self location exists
 		description: 'Project location must exist!'.
+	self 
+		assert: (IceRepositoryCreator isGitRoot: self location)
+		description: 'Project location does not seems to be a valid git repository'
 
 ]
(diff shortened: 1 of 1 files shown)


Iceberg-TipUI/IceTipLocalRepositoryPanel.class.st
```diff
@@ -86,11 +86,9 @@ IceTipLocalRepositoryPanel >> validate [
 
 	self
 		assert: self location isNotNil
-		description: 'Project location must exist!'.
+		description: 'Please provide a project location!'.
 	self
 		assert: self location exists
 		description: 'Project location must exist!'.
-	self
-		assert: (IceRepositoryCreator isGitRoot: self location)
```
jecisc, line 94: Is it normal that this assertion is removed?

AlexisCnockaert: sorry it was link to the fix of another issue mb

jecisc (approved):

PR #1874, opened 2025-01-06
renamed all ice git hub classes from model to presenter ( not model anymore)

AlexisCnockaert:
desc.


jecisc:
It would be nice to keep a deprecated alias. 

You can do it like this:

In IceGitHubAcceptPullRequestPresenter, class side you can add:

```st
IceGitHubAcceptPullRequestPresenter class>>initialize
    self deprecatedAliases: { #IceGitHubAcceptPullRequestModel }

```

This allows people referencing or extending those class to have a version of Pharo to migrate their code

Ducasse:
Thanks cyril 
We will have to bullet proof the aliases because they have some strange behavior.

PR #1875, opened 2025-01-09
Port Hernan's snapshot browser to Iceberg

jecisc:
https://github.com/pharo-project/pharo/pull/17468


Ducasse:
There is a problem with the baseline it cannot find 
Iceberg-TipUI-SnapshotBrowser-Tests

guillep:
Hi @Ducasse @jecisc this is not working, this morning @tesonep was trying to open a non-loaded package with this and he had an exception

PR #1887, opened 2025-01-28
Removing dead code + fix critique logic

Ducasse:
Cleaning IceTipCritiquesBeforeCommitBrowser logic.


Ducasse:
I need help @tesonep @jecisc because I do not get why the build is suddenly failing that much. 
I fixed a DNU due to the fact that the model of the critique was set after the instance creation. 
May be the tests are not in sync with the code of iceberg after my changes.

Ducasse:
Since the PR of seb fixed the problem. I close this one and I will harvest the changes.

PR #1894, opened 2025-02-11
Added a null presenter to push on the left. Remove dead code.

Ducasse:
Fix the cut plus in the branch from issue.


Ducasse:
totally unrelated.

PR #1896, opened 2025-02-15
Make #importOn: on IceGitChangeImporter take into account that the ‘fileReference’ is not necessarily a file when the size of the ‘path’ is 1

Rinzwind:
This pull request makes #importOn: on IceGitChangeImporter take into account that the ‘fileReference’ is not necessarily a file when the size of the ‘path’ is 1. Like [libgit2-pharo-bindings pull request #93](https://github.com/pharo-vcs/libgit2-pharo-bindings/pull/93), this avoids the signaling of a LGit_GIT_ENOTFOUND for [git submodules](https://git-scm.com/docs/gitsubmodules). For now, a submodule is just ignored and doesn’t appear in commit diffs.

### DIFF
--- a/Iceberg-Libgit/IceGitChangeImporter.class.st
+++ b/Iceberg-Libgit/IceGitChangeImporter.class.st
@@ -70,10 +70,10 @@ IceGitChangeImporter >> importOn: aNode [
 	For now, we only support packages inside the declared subdirectory (or root otherwise).
 	All other changes are simple directory/file changes"
 	
-	"If path = 1 this means it is a file"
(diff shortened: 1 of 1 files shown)


jecisc (approved):

Rinzwind:
@jecisc Thanks! Would you mind merging libgit2-pharo-bindings pull requests [#92](https://github.com/pharo-vcs/libgit2-pharo-bindings/pull/92) and [#93](https://github.com/pharo-vcs/libgit2-pharo-bindings/pull/93) as well?

PR #1899, opened 2025-03-06
Pharo13.0.0: #chore: Changed the automatic comment in commit message

AlexisCnockaert:
-Added the fixes# with issue number labels
-Put the current version of Pharo at the beggining of commit title, can be very useful if somebody searches a bug for example -Added all types of commit as comment, to have a clean way to work it would be better to use one of these types for each commit to detail what people have done.
@Ducasse or @jecisc let me know!

### DIFF
--- a/Iceberg-TipUI/IceTipCommentPanel.class.st
+++ b/Iceberg-TipUI/IceTipCommentPanel.class.st
@@ -61,8 +61,6 @@ IceTipCommentPanel >> defaultLayout [
 				   yourself)
 		  expand: false;
 		  add: commentText;
-		  add: fixesLabel expand: false;
-		  add: fixesField expand: false;
 		  add: actionButtonBar expand: false;
 		  yourself
 ]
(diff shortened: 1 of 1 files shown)


jecisc:
I don't have time to review but now that I think about it, I wonder if we should not have this message only for the first commit on a new branch?

Iceberg-TipUI/IceTipCommentPanel.class.st
```diff
@@ -121,10 +119,11 @@ IceTipCommentPanel >> initializePresenters [
 		               text:
 			               SystemVersion current majorMinorVersion , ':'
```
jecisc: I think it should start by two line returns to let the first line empty and let people add a title without the first thing we see been the Pharo version. 

When we do only 1 commit, it will be the automatic title of the PR and I'm not sure that the Pharo version is relevent for most projects as PR title

AlexisCnockaert: ok I understand the logic

Issue #1905, opened 2025-03-19
Enhancing IceGitLocal to support the '..' notation


tukanos:
Hi everyone,

I would like to use the RelativePaths '..' in the Metacello. In case we have local repositories that is handy way to indicate a parent.

Closely tied to the patch:
https://github.com/pharo-project/pharo/pull/18018/files

guillep:
Hi @tukanos, just for completeness, I put my comment here too.

Would you mind explaining what your issue is?
Do you have a reproducible issue? a stack trace?

Personally use relative paths at least containing '.' in gitlocal urls all the time...

For example, I have some scripts with

```smalltalk
Metacello new
  baseline: 'VMMaker';
  repository: 'gitlocal://./pharo-vm';
  load;
  lock.
Metacello new
  baseline: 'Ranger';
  repository: 'gitlocal://./ranger';
  load;
  lock.
```

PR #1906, opened 2025-03-19
Fixes: #1905 Enables local git repository to use '..' shortcut in path

tukanos:
The git local (gitlocal://) can now contain '..' shortcut.  The shortcut will be resolved with either workingDirectory path or a path which will be defined in the RelatviePath>>#expandedParent:

Closely connected to: https://github.com/pharo-project/pharo/pull/18018

### DIFF
--- a/Iceberg-Metacello-Integration/IceGitLocalRepositoryType.class.st
+++ b/Iceberg-Metacello-Integration/IceGitLocalRepositoryType.class.st
@@ -32,6 +32,16 @@ IceGitLocalRepositoryType class >> type [
 	^ 'gitlocal'
 ]
 
+{ #category : 'private' }
+IceGitLocalRepositoryType >> homoginizePath: urlPrefix [
+	"All possible paths are converted to absolute path"
+
+	| path |
+
+	path := Path from: (self location withoutPrefix: urlPrefix).
+	^ path asAbsolute asFileReference 
+]
+
 { #category : 'testing' }
 IceGitLocalRepositoryType >> isGitRoot: aReference [
 	^ IceRepositoryCreator isGitRoot: aReference
@@ -61,18 +71,21 @@ IceGitLocalRepositoryType >> mcRepository [
 
 { #category : 'private' }
 IceGitLocalRepositoryType >> splitRootAndSubdirectoryFromLocation [
-	| root subDir |
-	(self location beginsWith: (self class type, '://'))
-		ifFalse: [ self error: 'Invalid URL (It should be ', self class type, '://...)' ].
-	root := (self location allButFirst: self class type size + 3) asFileReference.
+	| urlPrefix  subDir root |
+	urlPrefix := self class type, '://'.
 	subDir := #().
+
+	(self location beginsWith: urlPrefix)
+		ifFalse: [ self error: 'Invalid URL (It should be ', urlPrefix ].
+	root := self homoginizePath: urlPrefix.
+
 	[ root isRoot or: [ self isGitRoot: root ] ] 
 	whileFalse: [ 
 		(root asAbsolute = FileSystem workingDirectory)
 			ifFalse: [ subDir := subDir copyWithFirst: root basename ].
 		root := root parent ].
 	root isRoot ifTrue: [ self error: 'I can''t find a .git/config file.' ].
-	
+
 	^ { 
 		root. "a FileReference"
 		String streamContents: [ :stream | subDir asStringOn: stream delimiter: '/' ] "a String" }


Ducasse:
Thanks!!!

Iceberg-Metacello-Integration/IceGitLocalRepositoryType.class.st
```diff
@@ -32,6 +32,16 @@ IceGitLocalRepositoryType class >> type [
...
 ]
 
+{ #category : 'private' }
+IceGitLocalRepositoryType >> homoginizePath: urlPrefix [
+	"All possible paths are converted to absolute path"
+
+	| path |
+
+	path := Path from: (self location withoutPrefix: urlPrefix).
+	^ path asAbsolute asFileReference
```
guillep, line 42: Following what I said in https://github.com/pharo-project/pharo/pull/18018

I strongly believe this change is not correct...

Iceberg-Metacello-Integration/IceGitLocalRepositoryType.class.st
```diff
@@ -61,18 +71,21 @@ IceGitLocalRepositoryType >> mcRepository [
...
+	root := self homoginizePath: urlPrefix.
+
 	[ root isRoot or: [ self isGitRoot: root ] ] 
 	whileFalse: [ 
 		(root asAbsolute = FileSystem workingDirectory)
 			ifFalse: [ subDir := subDir copyWithFirst: root basename ].
 		root := root parent ].
 	root isRoot ifTrue: [ self error: 'I can''t find a .git/config file.' ].
-	
+
```
guillep, line 88: This is super strange.

The code you're changed should already support '..' in github paths.

See for example:

```
'..' asFileReference
```

guillep (changes requested):
Hi @tukanos, would you mind explaining what your issue is? Because I personally use relative paths at least containing '.' in `gitlocal` urls all the time...

For example, I have some scripts with

```smalltalk
Metacello new
  baseline: 'VMMaker';
  repository: 'gitlocal://./pharo-vm';
  load;
  lock.
Metacello new
  baseline: 'Ranger';
  repository: 'gitlocal://./ranger';
  load;
  lock.
```

Do you have a reproducible issue? a stack trace?

Iceberg-Metacello-Integration/IceGitLocalRepositoryType.class.st
tukanos: I understand why you don't like it.  The issue here is when I can return path, but `#isGitRoot:` expects a reference.  If you have different idea, I'm opened to discussion.

```Smalltalk
isGitRoot: aReference
	^ IceRepositoryCreator isGitRoot: aReference
```	

The alternative would be:
```
[ self isGitRoot: root asFileReference ] ]
```

As for the motivation, relative path has a parent structure (except root path).   Relative path is, as you have written, specified by '..'.  This notation can be expanded 

 I would like to specify a parent of a relative path since Pharo is running in the docker image and then go to different directory tree of that parent.

The idea is to be able to move around the directory structure when you have changing absolute path, when docker is run on different computers, like this example in linux:

```
$pwd
/home/tukan/t 
$ls -la ../tmp
total 27208
drwxr-xr-x   4 tukan tukan     4096 pro 11 09:01 .
drwxr-xr-x 114 tukan tukan    12288 bře 19 14:31 ..
drwxr-xr-x   2 tukan tukan     4096 pro 11 09:01 backup
drwxrwxr-x   3 tukan tukan     4096 pro 11 09:00 lib
```

To take your example above:
```
Metacello new
  baseline: 'VMMaker';
  repository: 'gitlocal://./pharo-vm';
  load;
  lock.
```
This will take the current directory, but how do you go to a parent directory, when you would have a changing absolute path?

guillep: I understand how paths work, as well as I hope you do too :)

My point is: I believe that having relative paths in a gitlocal url in metacello is already supported.
We don't need a fix, because there is no bug.

But I may be wrong, so I'm just asking for an example that does not work that we can reproduce.

Iceberg-Metacello-Integration/IceGitLocalRepositoryType.class.st
tukanos: As I wrote above I would like to go to the parents directory.  As for the FileReference the reason was the `#isGitRoot` method.

I will demonstrate it on an example:

This is a working directory - which will change for every user (or any directory the user specifies in `#expandedParent`)
```
/home/<user>/<working_pharo_directory/
```

but e.g. libraries would be in different folders:
```
/home/<user>/important_stuff/libraries
```

so I would like to load the library like:
```
Metacello new
  baseline: 'VMMaker';
  repository: 'gitlocal://../important_stuff/libraries;
  load;
  lock.
```

If needed to go further you could (would have different directory structure):
```
Metacello new
  baseline: 'VMMaker';
  repository: 'gitlocal://../../important_stuff/libraries;
  load;
  lock.
```

guillep: I just did the following to load a (fresh) mutalk clone (from here https://github.com/pharo-contributions/mutalk) and this worked perfectly...

```smalltalk
Metacello new
  baseline: 'MuTalk';
  repository: 'gitlocal://../mutalk';
  load.
```

If you're having an issue there should be some difference in configuration.

I'm tested on: 

Pharo-13.0.0+SNAPSHOT.build.593.sha.9aa55368d349ba11822b37239dc8129f160b5fbb (64 Bit)
Build date: 19 March 2025

tukanos: Sigh, according to my testing you are right.  For whatever reason it works now when I use fresh image.  I would like to understand why it did not work for me before...  Thank you for you patience.

Iceberg-Metacello-Integration/IceGitLocalRepositoryType.class.st
tukanos: I undestand how paths works on most of the common systems.  On OpenVMS it is somewhat harder :).

PR #1917, opened 2025-03-28
RemoteModel should define repositoryModel (explicit trait requirement)

guillep:
~Just add a getter~

Actually, the issue is that the superclass defines the method.
But the subclass imports the trait and the `explicitRequirement` redefines the method.

I reposted a fix that only not imports the conflicting explicit requirement method.

Also, add a hash to the class that defines equals.

### DIFF
--- a/Iceberg-TipUI/IceTipRemoteModel.class.st
+++ b/Iceberg-TipUI/IceTipRemoteModel.class.st
@@ -30,6 +30,12 @@ IceTipRemoteModel >> description [
 	^ '{1} <{2}>' format: { self name. self entity url }
 ]
 
+{ #category : 'comparing' }
+IceTipRemoteModel >> hash [
+
+	^ entity hash
+]
+
 { #category : 'actions' }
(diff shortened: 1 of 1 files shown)


guillep:
After some analysis, the issue here is that  `explicitRequirement` has been deprecated!

```smalltalk
explicitRequirement
	"If one of the superclasses can perform the selector, we execute the method of that class, otherwise, the explicit requirement error is thrown"

	<debuggerCompleteToSender>
	| sender |
	
	self deprecated: 'This method will be removed in the future version of Pharo. See https://github.com/pharo-project/pharo/issues/15507 for details'.

	sender := thisContext sender.
	[ sender return: (self perform: sender selector withArguments: sender arguments inSuperclass: sender methodClass superclass) ]
		on: MessageNotUnderstood
		do: [ self error: 'Explicitly required method' ]
```

The thing is, for this kind of scenarios the deprecation gets called first, then the method in the superclass...
Of course, not importing the `explicitRequirement` method avoids calling the deprecation.

PR #1918, opened 2025-03-28
Introduce IceTipRepositoryStatusModel to avoid multiple commit lookups

guillep:
Opening iceberg does a lot of commit lookups to show info on the browsers.
This in turn stresses a lot the GC and the ephemeron queue because all those commits are set as autorelease.

This change adds a description object where values can be cached per table.
This allows to go from 1111 to ~33 lookups in my testing environment.


Ducasse:
@guillep what should we do? I can integrate this fix.

jecisc:
Some tests are failing but I think they are just failing in P13 branch and got fixed in P14.

Should we integrate and port forward to P14?

Ducasse:
@guillep @jecisc I would integrate this in P14. Is it ok?

jecisc:
Yes let's merge.
I was waiting to see if we could have another review because the Iceberg model is still a little obscure to me. But it would be a shame to lose an improvement like this :)

jecisc:
I have the impression that this broke a few things in Iceberg but I don't understand what is happening for now. I'll probably check this week

jecisc:
I found the error but I need to discuss with Esteban and Guille because I think this is a Spec bug but we need to decide what is the right behavior

jecisc:
I will revert this change for now because even with a fix in Spec, this is breaking the coloration in the repositories list

jecisc:
Reverted in: https://github.com/pharo-vcs/iceberg/commit/688b39d9d3f53428cfd13c58b8f3bce2f6a01e6c

PR #1923, opened 2025-04-08
New view for branches in Iceberg repositories

### LINKED ISSUE #1922: UX: Sort branches in the branches view
Image

In this view, branches seems to not be sorted. 

It might be nice to sort them by name.

Something even better would be to use a tree and group them by prefix before `/`.

Example:

Image

This is what does GitKraken

AlexisCnockaert:
- Display branches as a tree to have a better look, similar as Git Kraken or the ```tree``` command from Linux
Fixes: #1922 .
image


jecisc (changes requested):
- It would be nice to add an #expandAll 
- It would be nice to sort the items 
- The children need to be recursive `baseline/toto/tata`
- Disable OK button if we do not select a branch but a category

AlexisCnockaert:
Now looks like 
image

astares:
@AlexisCnockaert Can we add a (search) filter too (if the number of branches is high)

AlexisCnockaert:
@astares yes sure, no problem

jecisc (approved):

jecisc:
> @AlexisCnockaert Can we add a (search) filter too (if the number of branches is high)

In fact we enabled it but it does not appear. This seems to be a Spec bug

demarey:
Looks like there is a problem with this PR: I got a DNU when fixing a repo by checking out a branch.

Capture d’écran 2025-05-06 à 16 19 00

AlexisCnockaert:
Yes this PR need to make changes to methods calling the selectedBranch, I will fix this

AlexisCnockaert:
Fixed in #1931

PR #1929, opened 2025-05-05
Remove calls to a deprecated method

jecisc:
Found out that iceberg calls a lot a deprecated method while loading a project. This floods my logs.  Let's update it to not call it


jecisc:
I don't think the errors are related

PR #1930, opened 2025-05-06
Do not switch to commitish if commitish isNil

demarey:


demarey:
needs https://github.com/pharo-project/pharo/pull/18167

PR #1936, opened 2025-05-28
Added SyncToFiles command

jordanmontt:
## Preview

_Edit: updated image_

Capture d’écran 2025-05-29 à 14 23 55

## Description

Added new command named "SyncToFiles" that syncs the image changes to the github working directory. From the class comments:

> Syncs the image changes with the file system. Ensures that all changes to classes and methods are reflected in the Git repository’s working directory. It does not execute any staging (git add) nor commits.

## Motivation:

- Some times, especially with newcomers, we have a hard time setting up the ssh to properly work with git + iceberg. This new command allows to sync all the image changes to Git's working repository. With this, we can use the terminal or other git client to commit and push the changes.

- Another use case can be just to allow people to be able to use the git terminal client to do complex or non-conventional git magic.


jordanmontt:
Done with @Enzo-Demeulenaere during the Pharo sprint

Iceberg-TipUI/IceTipSyncToFilesCommand.class.st
```diff
@@ -0,0 +1,43 @@
...
+{ #category : 'default' }
+IceTipSyncToFilesCommand class >> defaultDescription [
+
+	^ 'Updates the file system to match the current state of the Pharo image. Ensures that all changes to classes and methods are reflected in the Git repository’s working directory, without staging or committing them.'
+]
+
+{ #category : 'default' }
+IceTipSyncToFilesCommand class >> defaultName [
+
+	^ 'SyncToFiles'
```
Gabriel-Darbord: I approve of this feature! :)
However, I have a few suggestions:
- All other buttons have a name written like a sentence, I think it'd be better if this one did too.
- I dislike the use of the word `sync` because that's usually associated with a bidirectional action, e.g. pulling and pushing. Here it's unidirectional, from the image to the file system, so how about using `write`, `export`, `update`, or yet another?
- I'm also unsure about the choice of icon, the floppy disk is tied to the idea of saving, but in a git context that idea would usually go with staging/committing in my opinion...
```suggestion
	^ 'Write to files'
```

Iceberg-TipUI/IceTipSyncToFilesCommand.class.st
```diff
@@ -0,0 +1,43 @@
...
+	IceTipSyncToFilesAction new
+		onSuccess: [
+			self inform: 'Image changes synched to Git''s working directory' ];
+		executeWithContext: context
+]
+
+{ #category : 'accessing' }
+IceTipSyncToFilesCommand >> iconName [
+
+	^ #smallSave
```
Gabriel-Darbord: How about `rightSide` instead? rightSide
It kind of looks like we're putting stuff in a file.
To be honest, I'm really unsure about which icon is best suited here.


```suggestion
	^ #rightSide
```

jordanmontt: Yes, I was also having a hard time choosing the icon.

Iceberg-TipUI/IceTipSyncToFilesCommand.class.st
jordanmontt: I agree that it should be a sentence. But I prefer to have the word "Sync" as I think it better describes what it does.

Iceberg-TipUI/IceTipSyncToFilesCommand.class.st
```diff
@@ -18,7 +18,7 @@ IceTipSyncToFilesCommand class >> defaultDescription [
 { #category : 'default' }
 IceTipSyncToFilesCommand class >> defaultName [
 
-	^ 'SyncToFiles'
+	^ 'Sync To Files'
```
Gabriel-Darbord: Sorry, I meant "written like a sentence" in the same sense as the other buttons, with spaces and where only the first word is capitalized.
For example, the button next to it in your screenshot is named "Merge to current".
image
That way, the UI keeps the same style.
```suggestion
	^ 'Sync to files'
```
Thanks!

Ducasse:
This is great to see more commands to help users.
Now I could not really understand what it was doing. Does it add changes from the image to the working directory or the working directory (unstagged files) to the image? Or both?

Iceberg-TipUI/IceTipSyncToFilesCommand.class.st
jordanmontt: ah yes!

jordanmontt:
Hello Stef @Ducasse 

It adds the changes from the image to the working directory. Not the unstagged files to the image.

Maybe we can improve the name/description ? Do you have any suggestions?

Gabriel-Darbord:
I'm sorry to keep pushing back on `Sync`.
I believe that when only reading the label, it makes us think that changes go both ways between the image and the working directory.

This is also the terminology used by VSCode source control.
Here are some excerpts from their docs where they talk about synchronization:
> VS Code offers you useful actions to push, pull, and [sync](https://code.visualstudio.com/docs/sourcecontrol/overview#_remotes) that branch (the latter will run a pull command followed by a push command).

> [Synchronize Changes](https://code.visualstudio.com/docs/sourcecontrol/overview#_git-status-bar-actions) will pull remote changes down to your local repository and then push local commits to the upstream branch.

I still think `Write` is the most direct and obvious term for what the action does.
If I were to express the action based on my understanding: it writes the current state of the classes in the image as files in the working directory.

Ducasse:
Yes Write looks better for what it does.

jordanmontt:
Ok! Let's go with write then

Gabriel-Darbord (approved):
Thanks, LGTM :)

PR #1940, opened 2025-06-03
Fix git direct access to repository

estebanlm:
while iceberg has always be capable of clone an arbitrary git repository (providing it has the right .git url), metacello, even meracello-iceberg-integration was not able to do it... thee was a mechanism implemented but it was broken. 

This PR fixes that (or tries :P).

Now you are capable of cloning a repository using this way: 

```smalltalk
Metacello new 
	baseline: 'CIG';
	repository: 'git:git@github.com:estebanlm/pharo-cig.git:main';
	load.
```

means, repository url is : 
1. `git:`  the  identifier.
2. `git@github.com:estebanlm/pharo-cig.git`  a git url like the ones provided by any git server nowadays. 
3. `:main` a branch

of course, this is providing that you have metacello-iceberg-integration enabled... which is the case in any pharo image nowadays)

### DIFF
--- a/BaselineOfIceberg/BaselineOfIceberg.class.st
+++ b/BaselineOfIceberg/BaselineOfIceberg.class.st
@@ -80,7 +80,7 @@ BaselineOfIceberg >> libgit: spec [
 		baseline: 'LibGit' 
 		with: [ 
 			spec
-				repository: 'github://pharo-vcs/libgit2-pharo-bindings:v3.2.0';
+				repository: 'github://pharo-vcs/libgit2-pharo-bindings:Pharo14';
   				loads: 'default' ].
 	spec  
 		project: 'LibGit-Tests'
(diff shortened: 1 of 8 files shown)


BaselineOfIceberg/BaselineOfIceberg.class.st
```diff
@@ -80,7 +80,7 @@ BaselineOfIceberg >> libgit: spec [
 		baseline: 'LibGit' 
 		with: [ 
 			spec
-				repository: 'github://pharo-vcs/libgit2-pharo-bindings:Pharo13';
+				repository: 'github://pharo-vcs/libgit2-pharo-bindings:v3.2.0';
```
jecisc: This should not be changed

BaselineOfIceberg/BaselineOfIceberg.class.st
```diff
@@ -132,5 +132,5 @@ BaselineOfIceberg >> resetKMRepository [
 BaselineOfIceberg >> tonel: spec [
 	spec
 		baseline: 'Tonel' 
-		with: [ spec repository: 'github://pharo-vcs/tonel:Pharo13' ]
+		with: [ spec repository: 'github://pharo-vcs/tonel:v1.2.0' ]
```
jecisc: Same here

estebanlm:
in fact it does needs to be changed, but to Pharo14 :P

jecisc (approved):

PR #1951, opened 2025-07-10
'Import from existing clone' exception handled by dialog

### LINKED ISSUE #18386: Iceberg raises exception on "Import from existing clone" for a fresh cloned image
Pharo 14 on Windows:
- clone a fresh Pharo 14 image with the launcher and start, 
- now open Iceberg and click on "Import from existing clone"
- Pharo gives Debugger because the "pharo-local\iceberg" does not yet exist (DirectoryDoesNotExistException
  (C:\Users\MyUser\Documents\Pharo\images\P14-New\pharo-local\iceberg in my case)

Image

AlexisCnockaert:
Entering a wrong path was raising an exception, which we don't want for the user. This is why in case this happens the dialog opens by default in the images directory, letting the user choose a repository to clone locally in other images (if he made a mistake or just typed random text ).
Fixes https://github.com/pharo-project/pharo/issues/18386

### DIFF
--- a/Iceberg-TipUI/IceTipDirectoryPresenter.class.st
+++ b/Iceberg-TipUI/IceTipDirectoryPresenter.class.st
@@ -14,7 +14,10 @@ IceTipDirectoryPresenter >> chooseReference [
 
 	| reference |
 	StOpenDirectoryPresenter new
-		openFolder: (self location ifNil: [ FileLocator workingDirectory ]);
+		openFolder:
+			((self location ifNil: [ FileLocator workingDirectory ]) exists
+				 ifTrue: [ self location ifNil: [ FileLocator workingDirectory ] ]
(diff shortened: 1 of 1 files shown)


Ducasse:
Tx alexis!

jordanmontt:
Sorry I did not see this before doing: https://github.com/pharo-vcs/iceberg/pull/1953

That solves the same issue. But, I do not understand the solution proposed here. It goes to fix the StOpenDirectoryPresenter instead of fixing faulty location.

I propose to use my PR. Because, it changes the default location to a location that will always exists. So, no need to touch the StOpenDirectoryPresenter presenter.

Ducasse:
@jordanmontt discuss this point with @guillep

jecisc:
Closing since it seems the issue has been fixed and accepted by Alexis

PR #1953, opened 2025-07-25
Using existing locations as default ones and tests

### LINKED ISSUE #18319: DirectoryDoesNotExist on Iceberg
**Bug description**

If you open Iceberg and try to add a new repository from local: `Import from existing clone` it fails when you open the FileBrowser because `DirectoryDoesNotExist`.

The default value is
```
/.../Documents/Pharo/images/p14-isloop/pharo-local/iceberg
```

But the `iceberg` folder does not exist (yes the pharo-local).

**To Reproduce**
1. Open Iceber view
2. Click on 'Add repository'
3. Select 'Import from existing clone'
4. Click on the small folder button

Image

**Expected behavior**
To not fail. 
In Pharo12 it opens the _last existing_ folder (`pharo-local` in the example).

We can do that, if not we can:
- change the default value in UI
- or create the missing folder on these cases (?

**Screenshots**

Image

**Version information:**
 - OS: MacOS
 - Pharo Version: 14.0

jordanmontt:
Fixes https://github.com/pharo-project/pharo/issues/18319

(First see issue's description)

This PR changes the default locations from selecting the repositories to a location that will **always** exists.
For example, before this PR, the repositoriesLocation tried to open the iceberg folder inside the Pharo image. BUT; this folder only exists in someone already cloned a repo in the same image. In a fresh image, this will fail with an exception as described in the issue.

It also adds two tests

### DIFF
--- a/Iceberg-Libgit/IceLibgitRepository.class.st
+++ b/Iceberg-Libgit/IceLibgitRepository.class.st
@@ -97,7 +97,7 @@ IceLibgitRepository class >> libgitRepositoryHandleAccessors [
 { #category : 'accessing' }
 IceLibgitRepository class >> localRepositoriesLocation [
 
-	^ FileLocator localDirectory
+	^ (FileLocator localDirectory / 'iceberg') ensureCreateDirectory
 ]
 
 { #category : 'accessing' }
--- a/Iceberg-Libgit/IceLibgitRepository.class.st
+++ b/Iceberg-Libgit/IceLibgitRepository.class.st
@@ -96,14 +96,16 @@ IceLibgitRepository class >> libgitRepositoryHandleAccessors [
 
 { #category : 'accessing' }
 IceLibgitRepository class >> localRepositoriesLocation [
-	^ FileLocator localDirectory / #iceberg
+
+	^ FileLocator localDirectory
 ]
 
 { #category : 'accessing' }
 IceLibgitRepository class >> repositoriesLocation [
- 	^ self shareRepositoriesBetweenImages 
-		ifTrue: [ self sharedRepositoriesLocation ]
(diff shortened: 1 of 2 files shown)


jordanmontt:
Also probably this deserves a backport to Pharo 13. Once it gets merged, I will do the backport too

jordanmontt:
It also fixes https://github.com/pharo-project/pharo/issues/18386

tesonep:
Why not adding an ensureCreateDirectory instead of changing the behaviour, I think a lot of people is assuming things with the path of the repository

jordanmontt:
I was thinking the same. 
- For the localRepositoriesLocation we can do it becauase anyways the iceberg folder will be created in the same pharo image.
- For the sharedRepositoriesLocation I do not know. Because we will be creating an iceberg folder in the HOME directory. and anyways that directory will be empty. And this is stored in a shared variable so it can be modified with the user preferences.

I think for case 1 we can do the ensureCreate but for the second one using home seems fine

demarey:
> I think for case 1 we can do the ensureCreate but for the second one using home seems fine

+1

jordanmontt:
I have updated this PR. It is ready for review and integration if the review passes.

- For the **local** repositories directory it will ensureCreateDirectory because it is the iceberg directory inside the pharo image. This cannot be modified 
- for the **shared** it just takes the HOME directory by default. It is a shared variable so the user can modify it

jordanmontt:
It will also close this issue: https://github.com/pharo-project/pharo/issues/18539

jecisc:
@AlexisCnockaert Since you've checked this issue, could you review this PR?

AlexisCnockaert (approved):
It looks all good

jecisc (approved):

PR #1954, opened 2025-07-25
Allow users to define a custom priority for their tag.

LeoDefossez:
Previously : custom tag = after "Projets" tab,  before "System" tab, sorted alphabetically.
Now : custom priority to define  more finely tabs order.
Hard coded dictionary moved to class side and new adding method #addTag: priority:


LeoDefossez:
Could this please be backported to pharo 13?

jecisc (approved):

PR #1968, opened 2025-09-19
Change ```IceTipDirectoryPresenter``` location condition

### LINKED ISSUE #18539: Error: Iceberg  directory does not exist when open github browser to choose a directory or file
when open github browser to choose a directory or file ,  open an Error pane because ceberg  directory does not exist 

StFileSystemModel >> currentDirectory: aFileReference

	(aFileReference exists and: [ aFileReference isDirectory ]) 
		ifFalse: [ DirectoryDoesNotExist signalWith: aFileReference ].
	currentDirectory := aFileReference.
	LastVisitedDirectory := currentDirectory

1\  Make sure  ...../pharo_local/iceberg    exist
or 2\  IceTipDirectoryPresenter >> chooseReference

	| reference |
	StOpenDirectoryPresenter new
		openFolder: (self location exist ifFalse: [ FileLocator workingDirectory ]); <------------- self location ifNil：[ ]       is not correct.
		title: self chooseTitle;
		okAction: [ :directoryReference | reference := directoryReference ];
		openModal.
	^ reference

IceTipLocationPresenter >> location
	"This method return the location to use in the repository
	If the location is relative to IceLibgitRepository >> #localRepositoriesLocation
	It returns a FileLocator so the image can be moved or the directory renamed.
	If the location is outside of the local repositories location it returns a file reference"

	| fileReferenceLocation relativePath|

	fileReferenceLocation := locationInput text trim
		ifNotEmpty: [ :value | value asFileReference ]
		ifEmpty: [ ^ nil ]. <-------------- FileLocator workingDirectory

	(fileReferenceLocation
		isContainedBy: IceLibgitRepository localRepositoriesLocation)
		ifFalse: [ ^ fileReferenceLocation ].

	relativePath := IceLibgitRepository localRepositoriesLocation makeRelative: locationInput text trim asFileReference.
	
	^ IceLibgitRepository localRepositoriesLocation resolvePath: relativePath

AlexisCnockaert:
To prevent an error while wanting to modify the path directly before opening the presenter, if the location doesn't exist it will open on a default location
Fixes https://github.com/pharo-project/pharo/issues/18539#event-19685907808

### DIFF
--- a/Iceberg-TipUI/IceTipDirectoryPresenter.class.st
+++ b/Iceberg-TipUI/IceTipDirectoryPresenter.class.st
@@ -14,7 +14,7 @@ IceTipDirectoryPresenter >> chooseReference [
 
 	| reference |
 	StOpenDirectoryPresenter new
-		openFolder: (self location ifNil: [ FileLocator workingDirectory ]);
+		openFolder: (self location exists ifFalse: [ FileLocator workingDirectory ]);
 		title: self chooseTitle;
 		okAction: [ :directoryReference | reference := directoryReference ];
 		openModal.


jordanmontt:
This is the same as this: https://github.com/pharo-vcs/iceberg/pull/1951

jordanmontt:
In my opinion, I could be wrong of course, this is not a solution for the problem because
- This fix goes to the file opener presenter, if the location is not existing, it will **always** open the working directory. If a users tries to open a non existing directory an exception should be raised. Here we are hiding the error by opening a non requested directory.
- The original issue is an iceberg problem so I think that the fix should go to iceberg.(See https://github.com/pharo-vcs/iceberg/pull/1953)

jecisc:
Closing since another PR got integrated

PR #1972, opened 2025-09-25
Edit metadata project confusion removed

### LINKED ISSUE #18519: setting the icerberg folder to .git/info leads to a mess
If in iceberg we select the .git folder instead of src or repository when we set the metadata,
Iceberg gets totally confused. We can commit but the code is not accessible.
It would be good to 
- forbid to use .git 
- create an src folder by default

AlexisCnockaert:
-Since there was no directory in a new repository, users used to select the ```.git``` directory, which leads to a mess for the repository
 -So there we ignore the git directory and create a ```src``` and select it by default
Fixes https://github.com/pharo-project/pharo/issues/18519#event-19841801219

### DIFF
--- a/Iceberg-TipUI/IceTipEditProjectDialogPresenter.class.st
+++ b/Iceberg-TipUI/IceTipEditProjectDialogPresenter.class.st
@@ -58,22 +58,15 @@ IceTipEditProjectDialogPresenter >> acceptCallback [
 { #category : 'actions' }
 IceTipEditProjectDialogPresenter >> addDirectory [
 
-	| newDirectoryName newDirectory previousSelection |
+	| newDirectoryName |
 	
 	newDirectoryName := self application newRequest
 		title: 'Create a directory';
 		label: 'Enter the new Directory name';
 		openModalWithParent: self window.
 	newDirectoryName ifNil: [ ^ self ].
+	self createNewDirectory: newDirectoryName .
 	
-	previousSelection := selectedDirectoryPath.
-	newDirectory := (previousSelection / newDirectoryName) asResolvedBy: self model fileSystem. 
-	
-	newDirectory exists 
-		ifFalse: [ newDirectory createDirectory ].
-	
-	sourceDirectoryTree roots: { self model fileSystem }.
-	self expandAndSelect: previousSelection
 ]
 
 { #category : 'initialization' }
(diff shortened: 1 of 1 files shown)


Iceberg-TipUI/IceTipEditProjectDialogPresenter.class.st
```diff
@@ -104,8 +97,20 @@ IceTipEditProjectDialogPresenter >> connectPresenters [
...
+	self expandAndSelect: (RelativePath with: 'src')
+]
+
+{ #category : 'creation' }
+IceTipEditProjectDialogPresenter >> createNewDirectory: aString [
+
+	| newDirectory previousSelection |
+	previousSelection := selectedDirectoryPath.
+	newDirectory := previousSelection / aString asResolvedBy: self model fileSystem.
+	newDirectory exists ifTrue: [ ^ self ].
```
jecisc:
```suggestion
	newDirectory ifPresent: [ ^ self ].
```

Iceberg-TipUI/IceTipEditProjectDialogPresenter.class.st
```diff
@@ -246,9 +246,8 @@ IceTipEditProjectDialogPresenter >> initializeDirectoryTree [
 						  evaluated: #basename;
 						  yourself));
 		roots: { self model fileSystem };
-		children: [ :each | each directories ];
-		whenSelectionChangedDo: [ :announcement | 
-			self sourceDirectorySelectionChanged: announcement selectedPaths ]
+		children: [ :each | each directories reject: [ :dir | dir basename endsWith: '.git' ] ];
```
jecisc:
```suggestion
		children: [ :each | each directories reject: [ :dir | dir basename = '.git' ] ];
```

jecisc: This should not happen much but we could have a file or directory ending with .git that is not the .git folder.

jecisc (commented):
I added just two little improvement suggestions

Iceberg-TipUI/IceTipEditProjectDialogPresenter.class.st
AlexisCnockaert: I understand it is just in case

Ducasse:
I love to see this happening. the default src is so great :)

PR #1973, opened 2025-09-25
Enable / disbale commit button while commiting something

### LINKED ISSUE #1969: Disable the commit button when we clicked on it
It happens often to me that I missclick and click two times on the commit button of the Iceberg browser when commiting changes.

Maybe we could make the button disabled when we click it and enable it again when the commit is done?

@AlexisCnockaert This is a UX problem so it can interest you

AlexisCnockaert:
Fixes #1969
The user could missclikc and click again on the commit button while a commit is being executed before :)


jecisc (approved):
Tried it and it works perfectly thanks!

PR #1975, opened 2025-09-26
Refresh Iceberg UI when switching tabs

### LINKED ISSUE #1955: Update iceberg UI when we load a project in a new tab
A project can define a special tab in Iceberg. But if you have Iceberg open and you load a project in a tab that does not exist, Iceberg UI is not updated and you need to close it and open it again to see your tab. 

Maybe we could refresh the UI when it happens to avoid this troublesome step.

AlexisCnockaert:
- Firstly when selecting ```Projects``` as tab it was duplicating the tab, fixed now :) -Then when changing the tab of a repository, user needed to close and open the iceberg browser to refresh the tabs
- Fixes #1955.

### DIFF
--- a/Iceberg-TipUI/IceTipRepositoriesBrowser.class.st
+++ b/Iceberg-TipUI/IceTipRepositoriesBrowser.class.st
@@ -279,17 +279,14 @@ IceTipRepositoriesBrowser >> subscribeToAnnouncements [
 { #category : 'initialization' }
 IceTipRepositoriesBrowser >> updatePresenter [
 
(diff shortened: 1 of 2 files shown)


AlexisCnockaert:
This PR needs https://github.com/pharo-spec/Spec/pull/1822 to be merged

Iceberg-TipUI/IceTipRepositoriesBrowser.class.st
AlexisCnockaert, line 1: I filter the pages because there was duplication for every group

PR #1981, opened 2025-10-07
P14-fix-metacello-checkout-of-gitlab-subprojects

tesonep:
- We need to support better URLs for the integration of Metacello request in private gitlabs with subprojects.
- Adding a test for the parsing
- Loading a package should first guarantee that the version is the correct one.


jecisc (approved):
This seems good but smalltalkCI is currently broken with P14. Should we merge?

PR #1989, opened 2025-11-08
Fixing the fetchAllRemotes DNU

Ducasse:
- some automated repair
- fix the fetchAllRemotes DNU

### DIFF
--- a/Iceberg-TipUI/IceTipRepairAction.class.st
+++ b/Iceberg-TipUI/IceTipRepairAction.class.st
@@ -18,7 +18,7 @@ IceTipRepairAction class >> allAvailableFor: aRepository [
 	^ ((self allSubclasses 
 		select: [ :each | each isAvailableFor: aRepository ])
 		sorted: [ :a :b | a order < b order ])
-		collect: #new
+		collect: [:aClass | aClass new ]
 ]
 
 { #category : 'accessing' }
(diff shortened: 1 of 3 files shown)


Ducasse:
```
ceberg-SmartUI not found
Iceberg-Pharo6 not found
Instance of MetacelloPlatform did not understand #withMetacelloLoadSessionDo:
MetacelloPlatform(Object)>>doesNotUnderstand: #withMetacelloLoadSessionDo:
MetacelloScriptApiExecutor>>execute
Metacello>>execu
```

jecisc:
The application has #informUser:during:

Maybe we should add an #informUserDuring: also letting the developper manage the label himself instead of providing a default label?

So first step is to add informUserDuring: in Spec then use it here

jecisc:
Hum... In fact we can directly use #informUser:during: I think. I'll propose a code update to fix the problem

Iceberg-TipUI/IceTipRepairAction.class.st
```diff
@@ -64,7 +64,9 @@ IceTipRepairAction >> executeWithContext: aContext model: aRepositoryModel [
 IceTipRepairAction >> fetchAllRemotes [
 	| remotes |
 	remotes := self repository remotes.
-	self application informUserDuring: [ :bar | 
+	self halt. "We should not use UIManager default.
+	Now the replacement is not clear to me. SD."
+	UIManager default informUserDuring: [ :bar | 
 		bar label: 'Fetching incoming commits'.
 		bar max: remotes size.
```
jecisc, line 71:
```suggestion
	self application informUser: 'Fetching incoming commits' during: [ :bar | 
		bar max: remotes size.
```

jecisc (commented):
I proposed a fix that should work I hope

jecisc:
I also did a PR to fix some of the CI's failures: https://github.com/pharo-vcs/iceberg/pull/1990

Ducasse:
Tx cyril. I found the fix for the unclosing tabs :) My mistake.

PR #1991, opened 2025-11-12
Fixing an halt. less uiManager

Ducasse:


Ducasse:
Strange error

Instance of TestSuite class did not understand #announcer
TestSuite class(Object)>>doesNotUnderstand: #announcer

PR #1998, opened 2025-12-04
Remove a few unused methods and replace a deprecation warning

jecisc:
This will allow to simplify a little more Monticello code afterward


jecisc:
Failure unrelated. Merging to enable more cleanings in Pharo repo

PR #1999, opened 2025-12-05
Remove the last informs going to UIManager in Iceberg

jecisc:
This change depends on a PR of Pharo: https://github.com/pharo-project/pharo/pull/19007


jecisc:
Since the related PR was merge and there are bad behaviors if this one does not get integrated, I'm integrating it

PR #2000, opened 2025-12-07
Improve diff performance by caching writerClass

stimberm:
The benefit of caching the writerClass is acknowledged in the comment of IceDiff>>#writerClass, but that implementation is not effective. The method #buildForChanges: will ultimately refer to the #writerClass of the source and target, and do so repeatedly, per changed file in the changes.

I made this change by profiling the following expression

```
myRepository branch diffTo: myRepository workingCopy
```

where the active branch of the repository is different from the reference commit of the working copy. There are roughly a thousand items in the collection passed to #buildForChanges:.

The duration drops from 9 to 6 seconds, more or less.

### DIFF
--- a/Iceberg-Libgit/IceGitCommit.class.st
+++ b/Iceberg-Libgit/IceGitCommit.class.st
@@ -7,7 +7,8 @@ Class {
 		'datetime',
 		'ancestorIds',
 		'comment',
-		'packageNamesCache'
+		'packageNamesCache',
+		'writerClass'
 	],
 	#category : 'Iceberg-Libgit-Core',
 	#package : 'Iceberg-Libgit',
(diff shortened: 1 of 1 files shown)


jecisc (approved):
Super nice!

I tried on my pharo repo. Time dropped from 18.5sec to 10.3sec :)

jecisc:
@stimberm If you are checking this part of the code maybe you can be interested in this change also: https://github.com/pharo-vcs/iceberg/pull/1918/

jecisc:
I integrated https://github.com/pharo-vcs/iceberg/pull/1918 and with this the same piece of code that was taking 18.5sec, then 10.3 with you change now dropped to 1.3sec!

PR #2001, opened 2025-12-07
[Pharo 13 backport] Backport speed up of Iceberg diff computation

jecisc:
I propose to backport the recent 2 PRs to speedup iceberg. We could merge it and release after a few days of testing the new Pharo14 image


jecisc:
Since one of the PR has problems, I propose to backport only one of the improvements

PR #2004, opened 2025-12-08
Clean some CI scripts

jecisc:
- Remove unused action
- Re-enable the tests-full
- Make actions be like actions of other external pharo repositories


jecisc:
This makes more tests run so I'll merge

tinchodias:
And... is same number of test packages being loaded? I am lazy to check it but i thought the pre post loading scripts were needed to load code

jecisc:
The postload and preload scripts are still here. I did not touch them. There are referenced in the .ston files :)

tinchodias:
ok! thanks

PR #2005, opened 2025-12-08
.ston files should not say they load #development branch

jecisc:
I think this setting is ignored but anyway we should not declare this.


jecisc:
This is weird... This causes the CI to not run any test... I don't understand why

jecisc:
Oh this is not the branch but the group. Sorry for the noise

PR #2013, opened 2026-01-16
choose of directory should rely on the application standard dialog (

estebanlm:
(it can be anything).
... and in any case you should not do just "new", that will not respect the POM (Presenter Object Model) :P

### DIFF
--- a/Iceberg-TipUI/IceTipDirectoryPresenter.class.st
+++ b/Iceberg-TipUI/IceTipDirectoryPresenter.class.st
@@ -11,12 +11,11 @@ Class {
 
 { #category : 'actions' }
 IceTipDirectoryPresenter >> chooseReference [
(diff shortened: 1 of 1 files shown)


estebanlm:
error not related (and I have no idea what it is about... did we enable strict mode?)

Ducasse:
No the strict mode is integrated but we will not set it up blindly.

Ducasse:
I opened an issue because we have an initialisation problem

https://github.com/pharo-vcs/iceberg/issues/2015

estebanlm:
now I have no idea @jecisc maybe you know what is going on ?

jecisc:
I fixed some things in the CI last week :P It now passes (except on windows but I don't know how to fix those tests :( )

PR #2025, opened 2026-03-18
Move extensions to base packages

guillep:
Follow up on https://github.com/pharo-project/pharo/pull/19389


guillep:
Tests are broken because it is impossible to move extension methods from Iceberg to Pharo with the state of affairs.

 - Pharo has the methods (scpUrl et al)
 - then it loads Iceberg, which overwrites those methods with an extension
 - then this CI process unloads Iceberg, removing the methods! But this does not put the overwritten methods back in place
 - the installation of new iceberg (this PR) does not have the extensions, as they are meant to be in Pharo :)

=> **Conclusion:** when the test runs the methods are not there

As such, I'll integrate this PR and fix any issue that appears (if any, but I'm positive that we will hit few or none) as fast as possible.

PR #2027, opened 2026-03-25
when collecting method history you just need one, not all parents

estebanlm:
while collecting the history, is enough to traverse just the first parent.
and it speeds up slightly the process.


estebanlm:
errors do not seems related at all

jecisc:
Maybe it can be nice to add a comment to explain why we don't need the other parents?

I have a hard time to understand why we should not check the commits of the branch been merged

estebanlm:
because you are revwalking (you are already traversing all commits) :P
let me see if I explain myself. As the revwalk traverses all commits, if you do not enter the second parent, it will anyway pass again for that second parent later, as normal part of the revwalk.

PR #2035, opened 2026-04-30
New button

### LINKED ISSUE #2034: Too much commits slow down the repository loading
When there is too much commits in a repository the loading can take a certain amount of time.
(see below a repo with many commits)

Image

The things to add would maybe be to limit the amount of commits shown and to have a button to increase this amount

Daxio0:
fix #2034 

Now, only the first 20 commits are shown at the beginning, and there is a button show 20 more : 
image

Its not finished : the button disappear after one use : 
image

### DIFF
--- a/Iceberg-TipUI/IceTipHistoryBrowser.class.st
+++ b/Iceberg-TipUI/IceTipHistoryBrowser.class.st
@@ -10,7 +10,8 @@ Class {
 	#instVars : [
 		'commitList',
 		'commitInfoTabs',
-		'loadSize'
+		'loadSize',
+		'allCommits'
 	],
 	#category : 'Iceberg-TipUI-View-Repository',
 	#package : 'Iceberg-TipUI',
@@ -150,8 +151,8 @@ IceTipHistoryBrowser >> refresh [
 { #category : 'accessing' }
 IceTipHistoryBrowser >> refreshCommitList [
 
-    | allCommits maxCommits |
-    allCommits := self model commitModels. 
+    | maxCommits |
+    allCommits ifNil: [ allCommits := self model commitModels]. 
     maxCommits := self loadSize min: allCommits size.
     commitList items: (allCommits first: maxCommits).
     commitInfoTabs selectedModel: ( allCommits ifNotEmpty: #first ifEmpty: [ nil ])
--- a/Iceberg-TipUI/IceTipHistoryBrowser.class.st
+++ b/Iceberg-TipUI/IceTipHistoryBrowser.class.st
@@ -125,7 +125,7 @@ IceTipHistoryBrowser >> loadAll [
 
 { #category : 'actions' }
 IceTipHistoryBrowser >> loadMore: anInteger [
-	self loadSize: self loadSize + anInteger.
+	self loadSize: ((self loadSize + anInteger) min: allCommits size).
    self refreshCommitList . 
 ]
 
--- a/Iceberg-TipUI/IceTipHistoryBrowser.class.st
+++ b/Iceberg-TipUI/IceTipHistoryBrowser.class.st
@@ -9,7 +9,8 @@ Class {
 	#superclass : 'IceTipBrowser',
 	#instVars : [
 		'commitList',
-		'commitInfoTabs'
+		'commitInfoTabs',
+		'loadSize'
 	],
 	#category : 'Iceberg-TipUI-View-Repository',
 	#package : 'Iceberg-TipUI',
@@ -107,11 +108,23 @@ IceTipHistoryBrowser >> initializeCommitList [
 
 { #category : 'initialization' }
 IceTipHistoryBrowser >> initializePresenters [
-
+	loadSize := 20.
 	commitList := self newTable.
 	commitInfoTabs := self instantiate: IceTipCommitInfoPresenter.
 	
 	self initializeCommitList
+]
+
+{ #category : 'accessing' }
+IceTipHistoryBrowser >> loadSize [
+
+	^ loadSize
+]
+
+{ #category : 'accessing' }
+IceTipHistoryBrowser >> loadSize: anObject [
+
+	loadSize := anObject
 ]
 
 { #category : 'accessing - ui' }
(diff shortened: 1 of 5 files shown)


TheSky599:
pair programming with me

Daxio0:
Now we have a load more button which work and doesn't disappear : 
Capture d’écran du 2026-05-29 13-59-36

Capture d’écran du 2026-05-29 13-59-50

And a load all button which allow you to have all commits shown : 

Capture d’écran du 2026-05-29 13-59-59

jecisc:
Maybe we can load more than 20 at first? 20 seems really not a lot to me. We should try with different values and see at what threshold it is starting to be slow.

Also, "Load more" could indicate the number of commits it will load with "Load 20 more". 
And last, when all commits are not loaded, it could be nice to have a label bellow the list of commit with a warning color saying "Only the latest X commits are displayed. Click on "Load 20 more" or "Load All" (This might take a while)"

Ducasse:
First let us integrate this then iterate.

Iceberg-TipUI/IceTipHistoryBrowser.class.st
```diff
@@ -107,13 +109,45 @@ IceTipHistoryBrowser >> initializeCommitList [
...
 
+{ #category : 'actions' }
+IceTipHistoryBrowser >> loadAll [
+	self loadSize: allCommits size.
+	self refreshCommitList .
+	
+]
+
+{ #category : 'as yet unclassified' }
+IceTipHistoryBrowser >> loadLess: anInteger [
```
Ducasse, line 127: I would rename loadLess: into: loadNumberOfCommits: or loadAmount:

Iceberg-TipUI/IceTipHistoryBrowser.class.st
```diff
@@ -107,13 +109,45 @@ IceTipHistoryBrowser >> initializeCommitList [
...
 	commitList := self newTable.
 	commitInfoTabs := self instantiate: IceTipCommitInfoPresenter.
 	
 	self initializeCommitList
 ]
 
+{ #category : 'actions' }
+IceTipHistoryBrowser >> loadAll [
+	self loadSize: allCommits size.
+	self refreshCommitList .
```
Ducasse: Can you remove the space and the period

Iceberg-TipUI/IceTipHistoryBrowser.class.st
```diff
@@ -107,13 +109,45 @@ IceTipHistoryBrowser >> initializeCommitList [
...
+{ #category : 'as yet unclassified' }
+IceTipHistoryBrowser >> loadLess: anInteger [
+
+	self loadSize: (self loadSize - anInteger max: 0).
+	self refreshCommitList
+]
+
+{ #category : 'actions' }
+IceTipHistoryBrowser >> loadMore: anInteger [
+	self loadSize: ((self loadSize + anInteger) min: allCommits size).
```
Ducasse, line 136: can tab consistently and remove the trailing space and .

Iceberg-TipUI/IceTipLoadMoreCommand.class.st
```diff
@@ -0,0 +1,24 @@
...
+{ #category : 'accessing - defaults' }
+IceTipLoadMoreCommand class >> defaultName [ 
+	^ 'Load More'
+]
+
+{ #category : 'executing' }
+IceTipLoadMoreCommand >> execute [
+	"Execute the actions that should be done by the command.
+	 This method expect that the context has been put in #context inst. var. if any context is relevant."
+	^ self context historyPanel loadMore:20.
```
Ducasse: a little space before 20

Iceberg-TipUI/IceTipRepositoryBrowser.class.st
```diff
@@ -22,7 +22,10 @@ IceTipRepositoryBrowser class >> buildGeneralCommandGroupWith: presenter for: aC
 		IceTipPullCommand.
 		IceTipFetchCommand.
 		IceTipCheckoutBranchCommand.
-		IceTipMergeCommand } do: [ :each | 
+		IceTipMergeCommand.
+		IceTipLoadMoreCommand.
+		IceTipLoadLessCommand .
```
Ducasse: remove the extra space please.

Ducasse:
This is a great feature. Thanks for your contribution. I did a little code review

Iceberg-TipUI/IceTipHistoryBrowser.class.st
```diff
@@ -137,11 +171,11 @@ IceTipHistoryBrowser >> refresh [
 { #category : 'accessing' }
 IceTipHistoryBrowser >> refreshCommitList [
 
-	| commitModels |
-	commitModels := self model commitModels.
-	commitList items: commitModels.
-	commitInfoTabs selectedModel: (commitModels ifNotEmpty: #first ifEmpty: [ nil ])
-
+    | maxCommits |
+    allCommits ifNil: [ allCommits := self model commitModels].
```
Ducasse: I was wondering why now we need the allCommits ifNil:....

Daxio0: The reason for the ifNil:, is because before ```allCommits := self model commitModels``` was called at each refresh, but since here when we refresh, we only change what is shown, we don't need to reload allCommits. Obviously rather than using a ifNil: for that, we could just load all commits before the first refresh only one time, but since the model need to have been set, i'm not really sure where to put it ? Because when i tried to do that instead : 
```smallTalk 
model: anObject
	super model: anObject.
	allCommits := self model commitModels.
	self refresh 
```
Then i have this error which means that allCommits is nil : 
image

Daxio0: Actually, it seems that i just needed to do that in the initializePresenters : 
```smallTalk
initializePresenters
	loadSize := 20.
	commitList := self newTable.
	commitInfoTabs := self instantiate: IceTipCommitInfoPresenter.
	allCommits := self model commitModels.
	self initializeCommitList
```

PR #2037, opened 2026-05-11
- Fixing deprecated code

Ducasse:
- RB to OC migration

### DIFF
--- a/Iceberg-TipUI-SnapshotBrowser/IceSBBrowseFullMethodCommand.class.st
+++ b/Iceberg-TipUI-SnapshotBrowser/IceSBBrowseFullMethodCommand.class.st
@@ -11,13 +11,13 @@ Class {
 { #category : 'default' }
 IceSBBrowseFullMethodCommand class >> defaultName [
 
-	^ 'Browse full'
+	^ 'Browse (in image) full'
 ]
 
 { #category : 'testing' }
 IceSBBrowseFullMethodCommand >> canBeExecuted [ 
 
-	^ super canBeExecuted and: [ self selectedClassIsLoaded ]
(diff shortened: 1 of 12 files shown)


Iceberg-TipUI-SnapshotBrowser/IceSBBrowseFullMethodCommand.class.st
```diff
@@ -11,13 +11,13 @@ Class {
...
 
-	^ 'Browse full'
+	^ 'Browse (in image) full'
 ]
 
 { #category : 'testing' }
 IceSBBrowseFullMethodCommand >> canBeExecuted [ 
 
-	^ super canBeExecuted and: [ self selectedClassIsLoaded ]
+	^   self selectedClassIsLoaded and: [ self selectedMethodDefinition isAlreadyDefined ]
```
jecisc, line 20: Are you sure we should not start with the super? It checks if the selected method is nil. If it is the second check here will crash

Ducasse: The super logic is not good. 
The problem is that some of these commands are for code inside the image (such as browse) and other (load) are too load code in the image (so check that the class is there)

```
Superclass>>selectedClassIsLoaded
	"Answer <true> if the currently selected class is loaded in the system"

	^ self class environment hasClassNamed: self selectedClass
```

```
Superclass>>canBeExecuted 

	^ self selectedMethod notNil
=> does not work because the class may not be loaded so selectedMethod does exist in the image
```

So I change it to 

```
Superclass>>canBeExecuted 
	"we cannot load a method of a class not loaded"
	
	^ self selectedMethodDefinition actualClass isNotNil
```

may be this is the same as selectedClassIsLoaded but I do not like the isNotNil because actual class looks magic to me.

PR #2040, opened 2026-05-29
Fix ssh error with passphrase

### LINKED ISSUE #2039: Issue ssh passphrase
I use a passphrase with my ssh key
When i have to clone one of my remote i have this error

Image

to fix the problem we need to remove this code : , self credentials hostDescription

AntoineUsal:
Fix #2039


AntoineUsal:
need to fix with the object hostDescription next time

PR #2044, opened 2026-06-11
no commits do not have tags

estebanlm:
this thing was causing me some infinite loop on certain cases


estebanlm:
crap, a problem in the PR

Ducasse:
@estebanlm did you see that when we want to add a tag using iceberg we got a DNU?

estebanlm:
nope, this was an infinite cycle, not error tagging

Ducasse:
Ok the error is then still there.

estebanlm:
I will check it

Ducasse:
Tx!

PR #2046, opened 2026-06-16
fix create tag on repositories browser

estebanlm:
Refreshing repositories because we may have some information to show.


estebanlm:
errors not related

PR #2050, opened 2026-06-26
Iceberg method modified with no change

Daxio0:
Before, if a method was modified with no change, the status in repositories was still put at Uncommitted changes


NathanMalenge:
https://github.com/user-attachments/assets/f70a812a-e9dd-4f2a-8448-6cd8b7782f32

Here is an exemple of how it worked. You can see that I recompile the method foo but the sourceCode does not change, however the repo status is `Uncommitted changes`

Daxio0:
https://github.com/user-attachments/assets/cd9ed8d8-f5f9-4c29-871c-b6ea5ceec8e1

Here is after the change, we can see that when there is a modification with no change, there is no `Uncommitted changes`

PR #2063, opened 2026-08-11
deepCopy to copy/postCopy

Ducasse:


Ducasse:
I will close this PR and redo it. 
- one for P15 
- and the astring for P15

